PR #1, opened 2016-01-03
Add version field

ivankravets:


marvinroger:
Thanks! But, from the docs: `version: Required if repository field is not defined`.
I assume this is to prepare the semver feature?

ivankravets:
Thanks! I've just improved documentation. Is it clearly now? http://docs.platformio.org/en/latest/librarymanager/config.html#version

marvinroger:
It is :smiley:

PR #6, opened 2016-01-17
Add mth to log current config

enavarro222:
It is not called in the code, it might be...

For now I use it to test/debug new config param (port, ota...)


marvinroger:
Closing in favor of #7

PR #7, opened 2016-01-17
Add MQTT port and OTA path & port to config

### LINKED ISSUE #3: Add config for MQTT port
To use it with a std MQTT broker it could be nice to make MQTT broker port configurable.

It's straightforward to add such option, however in a "pure" Homie framework it's just an other useless option... 

What do you think to add an `homie_port` option that take the value from the `HOMIE_PORT` constant if not given in the config JSON ?

I can propose a PR in this sense.

enavarro222:
This should close #3, this also include #6 but with to call to the log mth : 
* when config is saved in config mode
* at setup time in normal mode

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,11 +221,14 @@ Save the config to the device.
   "name": "kitchen-light",
   "wifi_ssid": "Network_1",
   "wifi_password": "I'm a Wi-Fi password!",
-  "homie_host": "192.168.1.10"
+  "homie_host": "192.168.1.10",
+  "homie_port": 35589,
+  "homie_ota_path": "/ota",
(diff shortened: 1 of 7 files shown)


marvinroger:
Sorry for the delay. Great! I only have a single objection: I don't like the fact the Wi-Fi password is exposed. Do you mind removing this sensitive information from the log method?

enavarro222:
Ok, I'll change that

enavarro222:
Done.

I just remove the password log for now.
Latter, t might be useful to indicate if there is a stored password or not...

marvinroger:
Awesome. Thanks!

PR #13, opened 2016-01-31
Change PIN_RESET to uint8_t and value to 0 (not D3)

enavarro222:
One line PR.

D3 (== GPIO0) is not defined when you are not using NodeMCU board in the arduino IDE, so you get a compilation error.

I guess float instead of uint was a copy past mistake...

Bytheway, I'm thinking to make this reset a bit more configurable... at least the PIN, but maybe also the "delay". An other way is to expose a function to manually trigger the reset from main code...


marvinroger:
Can you fill an issue about the configurable reset? Thanks!

PR #21, opened 2016-02-10
Added config.xml for phonegap build

furyfire:
I created a basic config.xml for Phonegap so it builds as an app.
You can test it here. (Android and Windows Phone only since I do not have a iOS developer license.)
[PhoneGap Build](https://build.phonegap.com/apps/1896165/share)


marvinroger:
Thanks! I'll CI-enable this soon.

PR #27, opened 2016-03-06
custom branding

vixns:
allow changing the "Homie" string of hostname / SSID

`Homie.setBrand("MyBrand");`


marvinroger:
Great, thanks!

Issue #42, opened 2016-03-21
TLS fingerprint checking?


jpmens:
As per [the JSON documentation](https://github.com/marvinroger/homie-esp8266/wiki/5.-JSON-configuration-file) I have uploaded a configuration which configures Homie to use TLS (the artist formerly known as SSL :-) and TLS connections are correctly established to a Mosquitto broker if I don't force the listener to `tlsv1.2`.

I _assumed_ the `fingerprint` element of the configuration would be to verify the server's certificate against some known good value. However, if I randomly modify the `fingerprint` on the device, the connection succeeds anyway. Is `fingerprint` checking actually implemented, or have I simply misunderstood the name of the element?

```json
{
  "name": "Clever name of device",
  "wifi": {
    "ssid": "yyyy",
    "password": "xxxx"
  },
  "mqtt": {
    "host": "192.168.1.130",
    "port": 9993,
    "ssl": true,
    "fingerprint": "99:07:4A:06:E7:EC:34:36:B7:F4:73:6C:CC:C3:49:60:14:F3:85:F3",
    "auth": false
  },
  "ota": {
    "enabled": true,
    "ssl": false,
    "host": "192.168.1.130",
    "port": 80,
    "path": "/ota"
  }
}
```
(The real broker's CA certificate fingerprint obtained with `openssl x509 -in ca.crt -noout -fingerprint` has a `2` at the end of the fingerprint instead of a `3`.)

jpmens:
Looking at [BootNormal](https://github.com/marvinroger/homie-esp8266/blob/3c96aa696b75c0f93819e585fb9fa30cbdeca260/src/Homie/Boot/BootNormal.cpp#L76) takes me [here](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/WiFiClientSecure.cpp#L407) and depths I wouldn't normally be in... it _appears_ to me as though it ought to work.

marvinroger:
As you just said, it is implemented and should work... Could you please add a `Logger.logln("Checking fingerprint")` at line 76? Just to be sure the condition is evaluated.

jpmens:
The condition is evaluated: I see my `logln()`, but I also see a bit of garbage: 

```
** Booting in normal mode **
⚙ Stored configuration:
  • Device ID: cffb1be0
  • Boot mode: normal
  • Name: Clever name of device
  • Wi-Fi
  • MQTT
    • Host: 192.168.1.130 
[...]

PR #43, opened 2016-03-21
Verify if fingerprint is NOT empty

jpmens:
I believe the logic is wrong: verification should be performed if the fingerprint string is not empty.
Addresses #42

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -73,7 +73,7 @@ void BootNormal::_mqttConnect() {
   }
 
   if (connectResult) {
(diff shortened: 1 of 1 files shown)


marvinroger:
strcmp returns 0 if string are equal, so you're right. I feel dumb!

marvinroger:
Thanks. :+1:

jpmens:
If it's any consolation, imagine how I feel after reading that one line 273 times this morning ... !

marvinroger:
I indeed feel better! Well, let's say we're not C++ guys, and we would expect from a string comparison function to return `true` if equals and `false` otherwise.

PR #84, opened 2016-05-11
Load the configuration UI into ESP8266 flash chip

unaiur:
In configuration mode, allow the webserver to serve any file in the flash prepending /http/ to the requested path.

Also, include any files from http://marvinroger.github.io/homie-esp8266/ that are needed to configure the device in data subdirectory. It's only 151KB, so it should fit perfectly in any flash size. I removed all the duplicated fonts, since woff format is supported by every browser nowadays.


marvinroger:
Awesome, so captive portal is enabled?

unaiur:
Yes, the captive portal is enabled if you put the contents of data
directory in the SPIFFS partition. Using PlatformIO, just copy that
directory next to your src directory and run the target "upload SPIFFS".

You need a 4Mbyte device. The font file alone is 90Kbyte and the 512Kbyte
devices only supports a 128KB filesystem.

If we could reduce that font leaving only the symbols actually  used, it
would fit in 512KB devices also.

Best regards!
On May 12, 2016 9:30 AM, "Marvin Roger" <notifications@github.com> wrote:

Awesome, so captive portal is enabled?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/marvinroger/homie-esp8266/pull/84#issuecomment-218681082>

marvinroger:
Well, to be honest I did not want the configuration UI to be embedded, because it requires another step (the upload to SPIFFS). I like your approach because even if the UI is not loaded, the API still works, so this is not a mandatory step.

I've managed to reduce both the icons and .js size, and to dynamically generate a single .gz html file that contains everything: http://marvinroger.github.io/homie-esp8266/ui_bundle.gz (33 KB)

This means we can simplify the code a lot, only serving `ui_bundle.gz` as `text/html` on `/`. :innocent:

unaiur:
I would keep the DNS interception code (all domain names are resolved to the local IP address) and handlers for "favicon.ico" & "/".
Are you going to accept it which these changes?

unaiur:
Ummm, I will try to embeed the favicon.ico into ui_bundle.gz to make it even easier. Good work!

marvinroger:
It is easy to embed in the HTML as a base64 string. :)
Of course I will merge this, this is a neat feature.

marvinroger:
41KB gzipped with the favicon embedded (link above updated).

unaiur:
Branch updated. I recompressed ui_bundle.gz with Zopfli to get a bit higher compression rate.

marvinroger:
I've once again reduced the resulting .gz to 33KB with some js optimization and using zopfli as you recommend. I'll merge this, but I will remove the .gz from the repo: the [configurator branch](https://github.com/marvinroger/homie-esp8266/tree/configurator) is automatically built on commits, and pushed to [gh-pages](https://github.com/marvinroger/homie-esp8266/tree/gh-pages). I don't want to copy/paste the .gz on the repo everytime the UI is updated. So I will add the link to the .gz file on the docs, so that user always download the up-to-date UI.

Thanks a lot for this! :+1:

PR #89, opened 2016-05-16
Added virtual methods to HomieNode

euphi:
Added virtual methods InputHandler(..), setup() and loop() to HomieNode, so derived class can overwrite them.

Changed Bootnormal to call setup() and loop() of each registered HomieNode.

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -127,10 +127,11 @@ void HomieClass::setBrand(const char* name) {
   strcpy(this->_interface.brand, name);
 }
 
-void HomieClass::registerNode(const HomieNode& node) {
+void HomieClass::registerNode(HomieNode& node) {
   this->_checkBeforeSetup(F("registerNode"));
   if (this->_interface.registeredNodesCount > MAX_REGISTERED_NODES_COUNT) {
     Serial.println(F("✖ register(): the max registered nodes count has been reached"));
+    Serial.flush();
     abort();
   }
 
--- a/src/Homie/Helpers.cpp
+++ b/src/Homie/Helpers.cpp
@@ -18,13 +18,13 @@ const char* Helpers::getDeviceId() {
 MdnsQueryResult Helpers::mdnsQuery(const char* service) {
   MdnsQueryResult result;
   result.success = false;
-  int n = MDNS.queryService(service, "tcp");
+  int n = 0; // MDNS.queryService(service, "tcp");
   if (n == 0) {
     return result;
   } else {
     result.success = true;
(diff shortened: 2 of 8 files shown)


src/HomieNode.hpp
```diff
@@ -15,12 +15,20 @@ class HomieNode {
...
   friend HomieInternals::BootNormal;
   friend HomieInternals::BootConfig;
-  public:
+
+  virtual bool InputHandler(String property, String value);
+  virtual void loop();
+  virtual void setup();
+
+protected:
+  bool _subscribeToAll;
```
euphi: it was necessary to make it protected (instead of private) to change it in constructor of derived Nodes.

src/Homie.cpp
```diff
@@ -127,10 +127,11 @@ void HomieClass::setBrand(const char* name) {
   strcpy(this->_interface.brand, name);
 }
 
-void HomieClass::registerNode(const HomieNode& node) {
+void HomieClass::registerNode(HomieNode& node) {
   this->_checkBeforeSetup(F("registerNode"));
   if (this->_interface.registeredNodesCount > MAX_REGISTERED_NODES_COUNT) {
     Serial.println(F("✖ register(): the max registered nodes count has been reached"));
+    Serial.flush();
```
euphi: Untested, but this should make the message visible on Serial.

src/Homie/Helpers.cpp
```diff
@@ -18,13 +18,13 @@ const char* Helpers::getDeviceId() {
 MdnsQueryResult Helpers::mdnsQuery(const char* service) {
   MdnsQueryResult result;
   result.success = false;
-  int n = MDNS.queryService(service, "tcp");
+  int n = 0; // MDNS.queryService(service, "tcp");
```
marvinroger: What's wrong?

euphi: it caused resets, may be releated to #79 / #81 .

unaiur: @euphi I fixed mDNS support in Homie and I've it working flawlessly in my lab for more than one week. Please, do not comment it. If you keep getting resets, raise a new bug report, please.

euphi: It was included in the pull request, because I just committed my local version based on v1.5.0.

I did not had the time to create a "clean" version, but wanted to show the implementation of the virtual methods, before I started a 5 day business trip.

As your pull request implements the same solution, I will close my pull request.

marvinroger:
Thanks for contributing @euphi :+1:

PR #90, opened 2016-05-20
Master onready

unaiur:
These are the changes I have pending in my repository. One allows to subclass HomieNode and get notified when the mqtt connection is ready, so it can publish additional properties for a node.

The other allows to publish arbitrary topics in MQTT. This allows me to build a switch that can publish any topic as true/false when touched.

Thank you for you great job. It has made mine much much smaller.


unaiur:
Merge request improved in #91

PR #91, opened 2016-05-20
Add several virtual methods to HomieNode

unaiur:
This is an alternate implementation of @euphi idea.
I've removed all the non essential editions (serial flush on some errors, disable mDNS) and I've added the onReadyToOperate() virtual method I proposed in my previous merge request.

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -127,7 +127,7 @@ void HomieClass::setBrand(const char* name) {
   strcpy(this->_interface.brand, name);
 }
 
-void HomieClass::registerNode(const HomieNode& node) {
+void HomieClass::registerNode(HomieNode& node) {
   this->_checkBeforeSetup(F("registerNode"));
   if (this->_interface.registeredNodesCount > MAX_REGISTERED_NODES_COUNT) {
     Serial.println(F("✖ register(): the max registered nodes count has been reached"));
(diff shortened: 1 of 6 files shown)


unaiur:
You can see a demo of the purpose of all these virtual functions in this project:
https://github.com/unaiur/homie-nodes

euphi:
I was also thinking about a way to "automagically" make the nodes register themselves, so you would only need to add the instance to your main.cpp and not make any extra call to any homie method.

First idea was to call `homie.registernode(*this)` in the constructor of `HomieNode`, but therefore it is necessary to ensure that `HomieClass Homie` is initialized first. C++ standard only ensures variable initialization order within the same compilation unit.

So, the solution would be to make `Homie.cpp` include a file like `Homienodes.hpp` that lists all Homienodes.

However, this solution would make it mandatory to create this file for any project and list all nodes there, while instantiating a node outside the file may lead to random resets.
Furthermore, you would need addtional forward declaration if you want to access a HomieNode object in some other compilation unit.

So, does someone have a better idea?

marvinroger:
Awesome ! Thanks to both of you.
@euphi could you please move the "automagically" discussion to the issue tracker?

unaiur:
We could call registerNode from HomieNode::setup()

euphi:
@unaiur : As proposed by @marvinroger I created a new issue #94 to continue discussion on virtual methods.

PR #92, opened 2016-05-20
Adds Homie::publishProperty

unaiur:
This method allow to publish properties for foreign objects.
For example, a switch can publish a property turning on a light.


marvinroger:
`path` and `property` could be simplified as a single `topic` parameter. And maybe renaming the method to `publishRaw(topic, payload)` would make it more obvious that this is not Homie relative. What do you think?

unaiur:
I have reworked the patch. I also added a method to get the topic associated to a homie node and made public the functions for reading the identifier of the homie device and homie nodes.

unaiur:
Reimplemented in #97

Issue #26, opened 2016-02-11
Interface for adding extra settings


marvinroger:
As @furyfire asked in #8, it would be neat to have an API to add other settings than wifi, mqtt and ota in `configuration` mode.

codmpm:
Hi Marvin, thanks for your obvious hard work.

I second the idea of having additional config parameters in configuration mode. Right now I'm building a NeoPixel Controller with mqtt connection and Homie-ESP just fits in to link this together. 
It would be very nice to configure the connected neopixel count or maybe the wanted protocoll for the neopixel stripe. Otherwise I have to put this in the sketch, which is ok but not as convenient as it could be if it was manageable from configuration mode.

[WiFiManager](https://github.com/tzapu/WiFiManager) for example has this possibility allthough you have to deal with the saving of the values by yourself. But it leaves me with the complete handling of events and mqtt-connection alone.

I've just wanted to give some insights why this option could be usefull. Thank you.

Regards,
Patrik

marvinroger:
Hi!

What do you mean by "the wanted protocoll"?

This is a feature I would like to implement at some point, but I need to work on the stability of the framework for the moment. There is a lot of new features to implement, and there are some heap issues showing up.

Anyway, thanks for telling your use case. ;)

codmpm:
Sorry, I should have clarified this:

When using neopixels you have to choose the right protocoll for the controller chip you are using. For exampl WS2811, WS2812(B), SK6812rgb, etc.
I try to build some kind of generic WiFi/MQTT/Neopixel Thingy.

I understand that stability os a bigger issue than additional parameters. Right now I can work with the mentioned WiFiManager and do the MQTT stuff myself. Thought the usecase could give you an idea :-)

You're welcome and thanky again.

marvinroger:
Alright, I got it. Let's discuss the implementation when the stability issues are resolved. :)

rschaten:
Like @codmpm, I started a project very similar to this, based on WiFiManager.

My plan is to place some nodes around the house, with different capabilities. Some of the nodes should get a DHT22 to measure temperature and humidity, some of my boards come with on-board RGB LED and a LDR, some should get relays to switch other devices.

My own firmware shows a form with some configurable values, via WiFiManager. I can configure the capabilities of this special node, and it announces this via MQTT (much like the `$nodes` in Homie).

[...]

PR #95, opened 2016-06-01
Transparent proxy for captive portal custom configurations

flaviostutz:
@marvinroger, as discussed in #26, I had implemented some apis for enabling more customized captive portals with minor impacts in base Homie firmware. With those apis in place, the developer can use the existing "ui_bundle.gz" capabilities to perform custom interactions with the user and/or cloud services.


marvinroger:
This is awesome! Thanks :+1:

PR #96, opened 2016-06-01
Proposal for autoregistration of HomieNodes

unaiur:
It removes the array of nodes from Datatypes/Interface.hpp and converts
it to a linked list. First & last pointers are stored as static members
of HomieNode.

Also, it adds 3 static methods to HomieNode: for_each() that executes a
lambda function on every registered node, find() which looks up a node,
and getNodeCount() which returns the number of registered nodes.


marvinroger:
:+1: thanks

Issue #99, opened 2016-06-08
Create a function in public Homie interface for deleting config.json


flaviostutz:
I need a way to wipe up device configuration (both to enable (mis)reconfiguration and/or change device owner). I thought of having Homie.eraseConfig() so that Homie would erase config.json. What do you think, @marvinroger? I would create a pull request...

marvinroger:
Merged in, thanks! :+1:

PR #100, opened 2016-06-08
Homie.eraseConfig() capability

flaviostutz:
Implementation proposal for discussion #99.


flaviostutz:
By the say, I just updated my branch and saw that you removed the Serial.begin(..) initialization from Homie: brilliant! I was in trouble with that because of a GPS module that talks in 9600... Thanks!

PR #144, opened 2016-09-02
Use new name of Json library

ivankravets:


marvinroger:
It was indeed pretty confusing, thanks.

ivankravets:
What do you mean under confusing? The previous name?

marvinroger:
Yes, well in other words it's a good thing it's now published under `ArduinoJson`.

ivankravets:
Aaa :) Cool. I don't remember why we named this library with `Json` previously.

PR #150, opened 2016-09-10
Topics/clough42/clear range

### LINKED ISSUE #149: Setting single node property re-uses range from previous set
This defect is against 2.0 (master)

If you set a node property using a range, like this:

    Homie
      .setNodeProperty(valvesNode, PROPERTY_TIME)
      .setRange(valve)
      .send(String(time));

and then later set a single (non-range) property, like this:

    Homie
      .setNodeProperty(batteryNode, PROPERTY_VOLTAGE)
      .send(String(voltageCoefficientSetting.get() * ESP.getVcc()));

The second set re-uses the left-over range value from the previous set.  In my case, the first set, above uses topic `devices/sprinklers/valves/time_6` and then the second set incorrectly sends `devices/sprinklers/powerUSB/voltage_6` when it should be just `devices/sprinklers/powerUSB/voltage`.  This is because HomieClass::setNodeProperty() doesn't clear the range when generating and returning the SendingPromise.

setNodeProperty() should clear the range in the sendingPromise before returning it.

I have a pull request for your consideration, coming shortly.

clough42:
For your consideration:

This change clears the range in HomieClass::setNodeProperty() before returning the SendingPromise to avoid contamination from the previous call.

I chose to add a clearRange() method to SendingPromise to facilitate this because the struct value is inconvenient to initialize for the existing setRange() method.

If you have another suggestion for how to accomplish this, I'll be happy to adjust.

This fixes #149.

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -22,10 +22,6 @@ SendingPromise& SendingPromise::setRange(HomieRange range) {
   _range = range;
 }
 
-SendingPromise& SendingPromise::clearRange() {
-  _range.isRange = false;
-}
-
 SendingPromise& SendingPromise::setRange(uint16_t rangeIndex) {
   HomieRange range;
   range.isRange = true;
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -22,6 +22,10 @@ SendingPromise& SendingPromise::setRange(HomieRange range) {
   _range = range;
 }
 
+SendingPromise& SendingPromise::clearRange() {
+  _range.isRange = false;
+}
+
 SendingPromise& SendingPromise::setRange(uint16_t rangeIndex) {
   HomieRange range;
   range.isRange = true;
--- a/src/Homie.hpp
+++ b/src/Homie.hpp
@@ -30,6 +30,7 @@ class SendingPromise {
   SendingPromise& setRetained(bool retained);
   SendingPromise& setRange(HomieRange range);
   SendingPromise& setRange(uint16_t rangeIndex);
+  SendingPromise& clearRange();
   void send(const String& value);
 
  private:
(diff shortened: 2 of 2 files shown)


clough42:
I see that the build failed (missing PubSubClient dependency), but I have no idea how I could have caused it.  It appears to be unrelated to my pull request, but if I caused it somehow, let me know how and I'll sort it out.

marvinroger:
The problem is `clearRange()` would be exposed to the sketch, which we don't want.
Actually there's a way to initialize a struct with the `setRange()`.

```c++
return _sendingPromise.setNode(node).setProperty(property).setQos(1).setRetained(true).setRange({ .isRange = false, .index = 0 });
```

Seems like PlatformIO was updated on the CI or something. @ivankravets sorry for the ping once again, what needs to be updated on our side?

ivankravets:
@marvinroger Thanks! Interesting moment. Looks like need to document it. PlatformIO 3.0 is really and has been significantly improved and simplified. No need to install library manually :) See my PR https://github.com/marvinroger/homie-esp8266/pull/151

----

Story...

PlatformIO handles dependencies when you install a library. It can be manual installation or specifying dependency in project file `platformio.ini`.  

Here we have 3-rd situation. Your library depends on a few other libs. It's OK and COOL. But... You use CI that doesn't handle dependency. You manually control which libraries to use and where they are located when call `pio ci...`. But... New PlatformIO 3.0 Build System is very smart. It found that project(1 sketch in your case) depends on Homie library and then it sees for nested dependencies. As result, PubSubClient library is required by isn't installed.

clough42:
@marvinroger Good call.  I wasn't really happy with that change.  I updated my PR branch with your suggested solution.  It's much better.

marvinroger:
Thanks to both of you, and kudos on pio 3! :)

PR #151, opened 2016-09-10
Install current library with all dependencies

ivankravets:


marvinroger:
Indeed, it looks and feels cleaner! Thanks :)

Issue #142, opened 2016-08-31
Feature Proposal: Listen to other topics -> Broadcast channel?


euphi:
As already proposed in #80 (but closed due to non-feasibility with the old PubSubClient), there should be an interface to subscribe to other devices/nodes.

My use case is to have a broadcast channel to send alarms and messages to all other devices.

In detail: I have a central OpenHAB server running with a local MQTT. The OpenHAB server should be able to publish MQTT message to all connected (Homie-)device, e.g. to trigger an alarm. I have some nodes with buzzer and display so they could show even some alarm messages.

You may even define a Broadcast channel in the Homie convention, e.g. `$broadcast/<level>` and `$alarm/<type>` ?
So, in my idea, any HomieNode shall be able to subscribe to broadcasts and alarms and gets called back.
Personally I would prefer an extra input-handlers e.g. `virtual void handleBroadcast(EBCstLevel level, String text)`.
**Update 2016-09-01:** So, in my current proposal there is no way for a Homie Node to directly publish on a broadcast channel.

If I get some feedback here, I will develop a proposed solution in my local fork.

marvinroger:
What about a way for you to subscribe to any topics you want? This way you can do pretty much everything. The current AsyncMqttClient implementation only allow one callback for messages, so you cannot override it, as Homie already handles messages. But I plan (https://github.com/marvinroger/async-mqtt-client/issues/21) to allow multiple callbacks. Would it be alright for you?

The broadcast channel is not a bad idea, but people will probably want more and more MQTT related-stuff anyway, so a complete access to the underlying MQTT client would solve all these problems.

euphi:
I think, broadcast channels should be part of Homie (the ESP implementation and the convention), but subscription to arbitrary topics should not. It's always good to have a quite strict API to force people to follow the convention.

So to allow arbitrary subscription, Homie should only offer a pointer (or reference) to the AsyncMqttClient instance, so anybody can bypass Homie and implement there own callback (as soon as 
https://github.com/marvinroger/async-mqtt-client/issues/21 is implemented).

I will be on a business trip and then on holidays during the next weeks. When I return I will develop my local fork to add support for a broadcast channel.

marvinroger:
The abitrary topics part would not be part of the convention, and yes to achieve arbitrary subs, you can use `Homie.getMqttClient()` which returns a reference to `AsyncMqttClient`.

Let's implement a broadcast channel, then!

marvinroger:
[...]

PR #161, opened 2016-09-29
#142: Add broadcast channel

euphi:
As proposed in #142 , I added a broadcast channel.

Sample implementation is at https://github.com/euphi/ESP_Heizungscontroller . (It does not make real use of it, it just logs the received message to Serial and back to MQTT).

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -100,6 +100,7 @@ HomieClass::HomieClass()
   _interface.reset.triggerTime = DEFAULT_RESET_TIME;
   _interface.reset.userFunction = []() { return false; };
   _interface.globalInputHandler = [](String node, String property, HomieRange range, String value) { return false; };
+  _interface.broadcastHandler = [](String level, String value) { return false; };
   _interface.setupFunction = []() {};
   _interface.loopFunction = []() {};
   _interface.eventHandler = [](HomieEvent event) {};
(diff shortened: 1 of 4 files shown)


src/Homie/Datatypes/Interface.hpp
```diff
@@ -37,6 +37,7 @@ struct Interface {
   } reset;
 
   GlobalInputHandler globalInputHandler;
+  BroadcastHandler broadcastHandler;
```
marvinroger: `BroadcastHandler` is defined nowhere

src/Homie.cpp
```diff
@@ -238,6 +239,14 @@ HomieClass& HomieClass::setGlobalInputHandler(GlobalInputHandler inputHandler) {
   return *this;
 }
 
+HomieClass& HomieClass::setBroadcastHandler(BroadcastHandler broadcastHandler) {
+  _checkBeforeSetup(F("setGlobalInputHandler"));
```
marvinroger: Can you replace this with `setBroadcastHandler`?

src/Homie/Datatypes/Interface.hpp
euphi: Seems I missed to add some changes for commit. 

I'm on vacation again, so I'll fix after October 20th.

src/Homie.cpp
euphi: Oh, of course.

euphi:
When I'm back, I'll merge your changes and create a new pull request

marvinroger:
Nevermind, I just fixed your PR. Thanks! 👍

PR #180, opened 2016-10-16
add iteadsonoff example that includes pushbutton handling

cpan-mila:
hi, this example is running on my devices and works great. this may also be useful for others.


shogsbro:
@cpan-mila it appears your PR does not include a direct changeset, but refers to a sub-project. I don't think @marvinroger will be able to review/merge your change.

cpan-mila:
ok, thanks. so let me try again...

PR #185, opened 2016-10-17
Subscribe ota even if disabled

mrpace2:
Can Homie subscribe to $ota even when $implementation/ota/enabled is false? That way, Homie could send a 403 ("FORBIDDEN") if someone tried updating firmware on an OTA-disabled device.


mrpace2:
Something went wrong with my merging. Closing this PR. Will redo.

PR #186, opened 2016-10-17
Subscribe ota even if disabled

mrpace2:
Can Homie subscribe to $ota even when $implementation/ota/enabled is false? That way, Homie could send a 403 ("FORBIDDEN") if someone tried updating firmware on an OTA-disabled device.

NB. Something is wrong with my forked repo. Even though (I think) I merged your develop branch into my develop, Github still says that I'm 8 commits ahead of you. I guess that is why this PR lists all my commits from the old PR, but only shows the changes made for this PR. Strange. Can I delete my forked repo and re-fork it without causing issues to the closed PR?


marvinroger:
No, you can refork and it won't harm the project, thanks for caring. I can still merge this PR as a single commit, don't worry. 😉

Issue #172, opened 2016-10-13
OTA: Verify firmware integrity before applying firmware update


mrpace2:
Homie devices should be better protected from bad data published to `$implementation/ota/payload` (intentionally or unintentionally).

Granted, there is some protection in place because we first need to post a version number to `$ota`. However, since `$ota` might be retained (the v2 spec allows retaining, *homie-ota* wisely doesn't), this protection could easily get lost. One bad publish to `$implementation/ota/payload` and you have to recover your homie thing from wherever it was installed, hook up a serial port, re-flash it and take it back to wherever it was.

The best solution I can imagine would be something along the lines of the flow chart on this website http://www.syncroness.com/portfolio/firmware-updates/ (found it using a quick web search, I'm not associated with that company). Unfortunately, public key infrastructure and RSA decryption might be a bit too heavy here. But, at least a hash or a checksum can easily be run over the firmware binary (that is, skip the key/RSA blocks in the picture). esp8266/Arduino Updater supports MD5 checking. We just need a Homie convention for OTA signatures and some code to enforce them.

For example, a Homie convention could request that the OTA server first publishes a firmware-hash like `md5:098f6bcd4621d373cade4e832627b4f6` to `$implementation/ota/signature` and then publishes the actual firmware to `$implementation/ota/payload`. By using an `md5:` prefix, the signature checking could later be extended to something more sophisticated. A new config key could specify whether or not a signature is required, and also which signature types are acceptable. That config key would need to be protected against lowering signature requirements.

Additional integrity tests could include

- matching the `Homie_setFirmware` name
- checking if the firmware size is within legal bounds

I can propose some code for the ESP side and instructions for the server side. Let me know.

flaviostutz:
+1

I was thinking about doing some sort of hash verification on my own... It would be great to be on the Homie itself.

marvinroger:
Definitely! Actually, I wanted to implement this, as as you said this is already implemented in the `Updater` class. But I forgot to. The `md5:` prefix could be removed because this is implementation related (see https://github.com/marvinroger/homie/issues/7), and MD5 is sufficient for our ESP8266, given the risk of collision is still very, very low.

I'm also :+1: for the `Homie_setFirmware` and firmware size bounds check!

mrpace2:
[...]

PR #189, opened 2016-10-18
Added OTA checksum verification

mrpace2:
See #172 

Here is a little test script for triggering OTA updates with a checksum:

    #!/bin/bash

    DEVICE=homie/awesome-relay
    BROKER=localhost
    FW_VER=1.0.0
    FW_BIN=/path/to/firmware.bin
    FW_MD5=$(md5sum -b $FW_BIN | cut -d" " -f1)

    mosquitto_pub -h $BROKER -t "$DEVICE/\$ota" -m "$FW_VER"
    sleep 1
    mosquitto_pub -h $BROKER -t "$DEVICE/\$implementation/ota/checksum" -m $FW_MD5
    sleep 1
    mosquitto_pub -h $BROKER -t "$DEVICE/\$implementation/ota/firmware" -f $FW_BIN

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -293,8 +293,7 @@ void BootNormal::_onMqttMessage(char* topic, char* payload, AsyncMqttClientMessa
   if (strcmp_P(device_topic, PSTR("$implementation/ota/firmware")) == 0) {  // If this is the OTA firmware
     if (!_interface->config->get().ota.enabled) {
       _publishOtaStatus(403);  // 403 Forbidden
-    }
-    else if (!_flaggedForOta) {
+    } else if (!_flaggedForOta) {
       _interface->logger->print(F("Receiving OTA firmware but not requested, skipping..."));
       _publishOtaStatus(400, PSTR("NOT_REQUESTED"));
     } else {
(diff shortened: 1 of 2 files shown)


marvinroger:
I would not unsubscribe from `firmware` and `checksum0`, otherwise you need, as you've done in your script, to wait a little bit before sending next messages. What do you think? Maybe you have a good reason to do so?

mrpace2:
Thanks for your feedback. Staying subscribed is actually what I was preferring when I was poking around in the code. I only left dynamic subscribe/unsubscribe in because that's what you had chosen for `payload` (now `firmware`) before. I was wondering why you had the `if (!_flaggedForOta)` check in the code. It should never have triggered because Homie only subscribed to the `payload` topic when `_flaggedForOta` was also true. Did you observe a race condition or something? Anyways. I added a commit to remove dynamic subscribe/unsubscribe.

BTW, the delay in my little test script is just for me to be able to observe what's going on looking at two terminals. I don't think the delay is needed. A real implementation would obviously evaluate Homie's status responses.

marvinroger:
Well we chose this because the payload was retained, so to receive it we had to subscribe only when receiving the `$ota` message. But at first the payload was not retained, and it looks like I forgot the flagged for OTA check. So yes, it was actually useless.

Great! LGTM 👍

PR #191, opened 2016-10-19
be more explicit about base64-padding

mrpace2:
- correct one leftover comment that @marvinroger noticed
- now checks firmware size when base64-padding is present


marvinroger:
Sorry about the lint 😅 but it actually helped to solve some tricky issues, as it checks more that only the code style.

Thanks, once again.

PR #195, opened 2016-10-23
Reject ota/firmware only once (vs. with every payload chunk)

mrpace2:
If an OTA server publishes firmware to `$implementation/ota/firmware`

- to a Homie device with OTA disabled, or 
- without having sent a new firmware version to `$ota` first

then Homie replies `403` (forbidden) or `400 NOT_REQUESTED` status for every chunk of payload it receives (which is about 350 times for a simple firmware).

This PR makes Homie reply only once per MQTT message.


mrpace2:
Thanks!

PR #196, opened 2016-10-23
Bugfix base64-encoded firmware when (size mod 3) equals 1

mrpace2:
I noticed an issue with base64-encoded OTA updates: for some firmware base-encoded OTA update worked and for some it didn't.

It turned out that, if the real firmware size modulo 3 equaled 1, firmware was incorrectly rejected with `400 BAD_FIRMWARE`. The reason was that the code I added only expected the real firmware size to be off by **one** from the "guestimated" size (3/4 ratio of the total length of the `=` padded base64 payload). However, the real firmware size may as well be off by **two** from the 3/4 ratio.

Examples with some realistic firmware sizes:

Real firmware size | Real size modulo 3 | Base64 length | Base64 Padding | Estimated firmware size | Size off-by
-------------------| ------------------ | ------------- | -------------- | ----------------------- | -----------
346992 | 0 | 462656 | (none) | 346992 | 0
346993 | 1 | 462660 | "==" | 346995 | 2
346994 | 2 | 462660 | "=" | 346995 | 1
346995 | 0 | 462660 | (none) | 346995 | 0
346996 | 1 | 462664 | "==" | 346998 | 2
346997 | 2 | 462664 | "=" | 346998 | 1
346998 | 0 | 462664 | (none) | 346998 | 0

Or, more general (N = any natural number):

Real firmware size | Real size modulo 3 | Base64 length | Base64 Padding | Estimated firmware size | Size off-by
-------------------| ------------------ | ------------- | -------------- | ----------------------- | -----------
N*3 | 0 | N*4 | (none) | N*3 | 0
N*3+1 | 1 | (N+1)*4 | "==" | N*3+1 | 2
N*3+2 | 2 | (N+1)*4 | "=" | N*3+1 | 1

This PR fixes the base64 length issue. In addition, the PR adds a check for payload size evenly dividable by 4 with base64-encoded payloads.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -18,7 +18,7 @@ BootNormal::BootNormal()
 , _otaChecksumSet(false)
 , _otaChecksum()
 , _otaIsBase64(false)
-, _otaBase64Padded(false)
+, _otaBase64Pads(0)
 , _otaSizeTotal(0)
 , _otaSizeDone(0) {
   _signalQualityTimer.setInterval(SIGNAL_QUALITY_SEND_INTERVAL);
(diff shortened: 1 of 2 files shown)


marvinroger:
Thanks a lot for your hard work on OTA!

mrpace2:
You're welcome! Glad if I can help.

PR #197, opened 2016-10-24
Make HomieNode CONST again! :-)

euphi:
Sending an MQTT message does not change the state of HomieNode so it should (and it can) be const.


euphi:
Sorry for the "8 commits", I messed up my github fork. However, at the end, there are only 2 lines changed (as I did in my local branach).

marvinroger:
No problems, commits are squashed into a single one anyway :)
Thanks!

PR #216, opened 2016-11-12
Add mac info message on boot normal

AlexValois:
When several devices are used and several are waiting their configuration, it is useful to check the mac address in Homie-ota server. Thanks to this change, the mac appear in inventory of the device in homie-ota server.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -163,7 +163,7 @@ void BootNormal::_onMqttConnected() {
 
   Interface::get().getMqttClient().publish(_prefixMqttTopic(PSTR("/$homie")), 1, true, HOMIE_VERSION);
   Interface::get().getMqttClient().publish(_prefixMqttTopic(PSTR("/$implementation")), 1, true, "esp8266");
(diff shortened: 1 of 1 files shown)


marvinroger:
The `DeviceId::get()` is not actually the mac address, but something made up of the ESP8266 and flash chip IDs. So `$mac` is probably bad named.

AlexValois:
Indeed, this is only the last half of the mac to which 0e is added. I will rename it.

thanks

marvinroger:
I'll merge this, but thinking about it `deviceId` is kind of confusing, `mac` is probably better, even though it's not actually a MAC address.

jpmens:
`hwaddr` (Hardware address) ?

Nzbuu:
There should be a real (station) MAC address available somewhere. In
Nodemcu, it's wifi.sta.getmac().

marvinroger:
I am considering replacing the hardware device ID (which is, as said above, something made up of the ESP8266 and flash chip IDs) by the complete 12 characters MAC address. Read https://github.com/esp8266/Arduino/issues/2309 for the reason why, but TL; DR:

> @igrr: There are now more OTP bytes used to set MAC address because we have run out of space with three bytes. getChipID doesn't take these new bytes into account, it has to be updated.
Edit: using MAC address as unique ID a good approach for now.

PR #218, opened 2016-11-15
bugfix: wrong buffer used when logging received ota checksum

mrpace2:
I noticed unterminated-string garbage in the log output... Probably not related to the open OTA issues.


marvinroger:
Right, thanks!

PR #219, opened 2016-11-15
Bugfix: base64 decoding

### LINKED ISSUE #208: OTA with BAD_CHECKSUM making device not responsible
Hi,

I've been testing the new changes with OTA and base64 and I haven't been able to successfully get it to work yet. 

To report the current situation, when sending a ota (using @mrpace2 script) the device is issuing a 400 BAD_CHECKSUM, but rebooting and then not responding until hard reset.

I can provide the logs from serial and mqtt if needed.

Is there anyone working on this? Otherwise I can take a look at it and try to solve it.

mrpace2:
Fixed an issue decoding base64. The first base64 characters of consecutive payload chunks were not handled correctly. The first two base64 characters matter, not just the first one.

This should fix #208.


marvinroger:
Thanks a lot! 👍

Issue #220, opened 2016-11-15
Proposal: replace $ota/version with $implementation/ota/checksum


mrpace2:
I think md5 hashes should be mandatory with Homie OTA because they are a good safety net for different kinds of potential OTA issues (see #204, #208). In addition, an md5 hash could eliminate the need for constantly changing an "artificial" firmware version during development (and with it, the inconsistency with the `$ota` topic :-)). I suggest keeping `$fw/version`, but just for informational purpose.

A new flow could be like this:

- During startup, Homie reports the current firmware's md5 to `$fw/checksum` (in addition to `$fw/name` and `$fw/version`). OTA servers may or may not use this information for automatically scheduling OTA updates.
- OTA update:
  - OTA server sends the new checksum to `$implementation/ota/checksum`. Homie compares its own md5 hash with new value and responds:
    - `403` Forbidden (if OTA disabled)
    - `304` Not Modified (if new md5 equals that of the running firmware)
    - `202` Accepted (otherwise)
  - For `202`, OTA server sends new firmware blob to `$implementation/ota/firmware`. Either base64-encoded or binary. Homie updates firmware and reboots. Same as today.

I could volunteer for implementing this, and I could also look into the matching `homie-ota` PR.

jpmens:
I would welcome this, in particular for `eliminate the need for constantly changing an "artificial" firmware version during development`.

marvinroger:
I agree, but how do you store/calculate the current firmware MD5? We can store it after a first OTA, but what happens for the very first OTA after a Serial flash?

mrpace2:
`ESP::getSketchMD5()`. It's the exact same hash as what we use during download. I checked.

marvinroger:
Awesome! Then yes, let's implement this 😇

marvinroger:
Closing this as this is now implemented and the [docs are updated](https://homie-esp8266.readme.io/v2.0.0/docs/ota-configuration-updates).

PR #221, opened 2016-11-15
Patch replace ota version with checksum

mrpace2:
Implements #220 

Note that, if OTA fails, we don't need to restart anymore just for resetting the expected MD5 because now there WILL be another SetMD5() when the server retries the update.

Updated bash script:

    #!/bin/bash

    DEVICE=homie/<device>
    BROKER="-h <broker>"
    FW_BIN=/path/to/firmware.bin

    FW_MD5=$(md5sum -b $FW_BIN | cut -d" " -f1)

    FW_B64=${FW_BIN/.bin/.b64}
    base64 -w0 $FW_BIN > $FW_B64

    mosquitto_pub $BROKER -t "$DEVICE/\$implementation/ota/checksum" -m $FW_MD5
    mosquitto_pub $BROKER -t "$DEVICE/\$implementation/ota/firmware" -f $FW_B64

I added a second commit to be slightly more consistent with icons and OTA message grammar.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -53,7 +53,7 @@ uint16_t BootNormal::_publishOtaStatus_P(int status, PGM_P info) {
 
 void BootNormal::_endOtaUpdate(bool success, uint8_t update_error) {
   if (success) {
-    Interface::get().getLogger() << F("✔ OTA success") << endl;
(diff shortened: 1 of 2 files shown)


mrpace2:
😅 Hooray! Lint passed straight away.

marvinroger:
Good job! And congrats for the lint, you're becoming a master 😄

Issue #204, opened 2016-10-31
MQTT OTA csum error in reboot after "OTA sucess"


Gulaschcowboy:
Hi,

with latest commits of homie-esp8266 I get the following result when updating my ESP8266-07s using homie-ota:

✴ OTA available (version 1.0.11)
OTA started
Triggering OTA_STARTED event...
Firmware is base64-encoded

<snip>

Receiving OTA firmware (353600/353600)...
✔ OTA success
Triggering OTA_SUCCESSFUL event...
Triggering MQTT_PACKET_ACKNOWLEDGED event (packetId 51)...

<snip>

Device is idle
↻ Rebooting...
Triggering MQTT_PACKET_ACKNOWLEDGED event (packetId 375)...
��� MQTT disconnected
Triggering MQTT_DISCONNECTED event...
↕ Attempting to connect to MQTT...
✖ Wi-Fi disconnected
Triggering WIFI_DISCONNECTED event...
↕ Attempting to connect to Wi-Fi...

 ets Jan  8 2013,rst cause:1, boot mode:(3,6)

load 0x4010f000, len 1384, room 16 
tail 8
chksum 0x2d
csum 0x2d
v3de0c112
@cp:0
ld

 ets Jan  8 2013,rst cause:4, boot mode:(3,6)

wdt reset
load 0x4010f000, len 1384, room 16 
tail 8
chksum 0x1d
csum 0x1d
csum err

Flashing same binary using serial works.
Any ideas?

Thanks,
Alex

clough42:
I am experiencing something similar.  I can get the OTA to upload to the device, and it apparently succeeds, but the device fails to reboot.  If I rest the device, I get only the off-baud garbage string and the device doesn't come up.

If I flash the same image via serial, it works.

I am building with PlatformIO, if that makes a difference.  I have the latest code as of today from the develop branch and the latest versions of the dependencies, from their respective repositories.

Here's my output, for reference.  [...snip...] is where I took out long sequences of similar messages:

```
`** Booting into normal mode **
{} Stored configuration
  • Hardware device ID: d1008cef
  • Device ID: firelight1
  • Boot mode: normal
  • Name: FireLight
  • Wi-Fi:
    ◦ SSID: clough
    ◦ Password not shown
  • MQTT:
    ◦ Host: mqtt
    ◦ Port: 1883
    ◦ Base topic: devices/
    ◦ Auth? no
  • OTA:
    ◦ Enabled? yes
[...]

PR #225, opened 2016-11-16
Removed String class usage in _publishOtaStatus, trying to workaround issue 204

mrpace2:


mrpace2:
@marvinroger Something is wrong with Travis. Can you please check? I added a 2nd commit because for the first commit Travis timed out waiting for two build jobs that never launched. The same thing just happened again for the second Travis run.

Finally, I was able to duplicate the exception on my setup. It triggered when I ran the Homie app that was built on Linux (latest Arduino 1.6.12). This is strange. It never crashed for me when built under Windows either with Arduino or with PlatformIo. 

Anyway. I confirmed that the commit from this PR fixes the exception in the Linux/Arduino case. I had also tested it on Windows - it worked. It worked on all tested platforms, both for binary and base64 firmware. As far as I can tell, this PR fixes the issue.

For curiosity, I will also try out PlatformIo on Linux when I get a chance (probably not today).

marvinroger:
@mrpace2 something went wrong on Travis, indeed, but it seems fixed now.
Let's wait for answers on #204 before merging this.

mrpace2:
Sure. I would also like to hear from other users who reported problems (@ckrey, @clough42, ...?) if it works now. Thanks in advance.

Testers may simply use the <https://github.com/mrpace2/homie-esp8266/tree/patch-trying-to-fix-issue-204-v2> branch from my fork for this test. The `patch-trying-to-fix-issue-204-v2` branch contains  the latest `delelop` code from the upstream repo, plus the changes from this PR that (I believe) fixes #204.

mrpace2:
@marvinroger Linux/PlatformIo also crashed before I applied the fix. With the fix applied, it works on my setup.

Gulaschcowboy:
@marvinroger @mrpace2 
Hi, I did some tests right now, with the patched library of @mrpace2 :

With base64:
- OTA with same hash gets sucessfully rejected, no crash
- OTA with different hash starts update, crashes at the end, see .txt

[base64-v100-v101-mrpace-patch.txt](https://github.com/marvinroger/homie-esp8266/files/598425/base64-v100-v101-mrpace-patch.txt)

Switching homie-ota to b64 to false works, see .txt
[NO-base64-v100-v101-mrpace-patch.txt](https://github.com/marvinroger/homie-esp8266/files/598437/NO-base64-v100-v101-mrpace-patch.txt)

mrpace2:
@Gulaschcowboy I fixed a type cast issue with OTA status logging that only matters with certain build settings in my fork (same branch, commit  b182d7e). Can you try again? Also, I'd like to understand why you keep getting different results. You're on Linux/Arduino, right? Can you please post your compilation output from the Arduino bottom panel with verbose output enabled (File > Preferences > check Show verbose output during compilation)? Thanks.

Gulaschcowboy:
@mrpace2 Attached a verbose compile with Arduino 1.6.12 on openSuse 42.2 against your branch before commit b182d7e. 
I have to check commit b182d7e later, no time today.

[verbose-compile.txt](https://github.com/marvinroger/homie-esp8266/files/599226/verbose-compile.txt)

mrpace2:
Thanks. Exact same compiler and linker options here.

mrpace2:
Because `info` is now a `PGM_P`.

Gulaschcowboy:
@mrpace2 Tried commit b182d7e now:

b64 false:
same as before
hash to same hash -> not requested, no crash ->OK
hash to different hash -> Update, no crash -> OK

b64 true (same with homie-ota and your script):
hash to same hash -> not requested, no crash ->OK
hash to different hash -> tries to update, aborts with bad checksum, continues on old firmware

So at least its doesn't crash anymore, but it detects a wrong hash, but it should be ok, see txt

[b64-after-b182d7e.txt](https://github.com/marvinroger/homie-esp8266/files/601362/b64-after-b182d7e.txt)

Gulaschcowboy:
@mrpace2 @marvinroger 
Hi Guys,

any progress on this? See my last test

mrpace2:
@marvinroger Are you planning to merge this PR or would you prefer if I withdrew/closed it? I would be fine with either decision because it worked for me, even without this PR and I can't see anything wrong with the code that used the String class.

The few responses to my request for test above indicate that, even though it's not fully understood why it crashed on some setups, at least, it's not crashing those anymore. I guess more feedback could be collected if this code was in your repo vs. users having to checkout a specific branch from my fork.

mrpace2:
@Gulaschcowboy Sorry, I currently don't have bandwidth for remotely troubleshooting the issue on your setup. You may use binary FOTA if - for whatever reason - base64 doesn't work for you. Or, dig into the code and contribute a bug fix if you find something.

Gulaschcowboy:
@mrpace2 Hi, sorry, but my coding skills are not enough to analyze what is going wrong about the checksum check. So I can't provide a patch, I can only provide live testing.

Binary OTA works, so I can live with that, while I think it is still interesting, why I should be the only one, that has this issue. Btw.: I have this with Arduino IDE and platfromio. 
So for you it works with b64 and correct checksum? What do you use to compile? Windows?

marvinroger:
@mrpace2 I am uncomfortable with the idea of fixing something that *should* work with the current code. There's obvously something wrong in either `esp8266/Arduino` or the `xtensa` toolchain, the problem does not come from this project. The problem is, the issue comes from the `String` class, and it's used in a lot of places in the code, so the issue we try to workaround here might happen again. If we're able to determinate what's exactly the problem, we can avoid that. But don't close the PR.

`esp8266/Arduino` seems dead ATM, so it's not the right time to open an issue there.

I have no problems with sketches compiled on the Arduino IDE on Windows. Let's say the issue is CPU related. Maybe the Windows xtensa is better at optimization than the Linux one. Can you try with a CPU frequency of 160MHz? (Tools -> CPU Frequency)

mrpace2:
I totally agree with your hesitation to workaround something that doesn't look like it's broken. I won't close this PR.

I am currently busy with something else and I won't be able to contribute to this project for a while. Can I delete my `homie-esp8266` fork or do I need to keep it and the `patch-trying-to-fix-issue-204-v2` branch because of this stale PR?

marvinroger:
@mrpace2 no problems, thanks a lot for what you've done. 😇
I have no idea! Not sure how GitHub handles that.

mrpace2:
@marvinroger Thank YOU for the **fantastic job** you're doing with `Homie` and `homie-esp8266`!

If you don't mind I'd like to try out if GH handles this. If this PR gets broken because of me deleting my fork, I will create a new PR, squashed and rebased to the HEAD of `develop`.

shogsbro:
I think you need to keep the branch from which you sourced the PR, at least that's what I've always done, but will await your report with interest :-)

euphi:
obsolete

PR #228, opened 2016-11-19
Clearing config/set topic after applying patch to avoid endless loops when using retained messages

flaviostutz:
Just got an endless loop when I used a retained message for patching Homie configuration. Clearing the message (publish an "" to the topic) before rebooting will avoid this.


marvinroger:
Thanks!

PR #232, opened 2016-11-20
Addition of IP info on console during SoftAP startup

flaviostutz:


flaviostutz:
Just added a commit removing the default password of SoftAP for usability improvements. After DeviceID has changed to the full mac address, it was hard for the user to memorise it during first AP connection.

In most cases an unprotected AP won't be much dangerous because it will be active only during the first steps. On more robust systems, the developer could put a default password on config.json.

What do you think?

marvinroger:
Thanks!

PR #240, opened 2016-11-24
Simplification on boot modes logic. Now any mode can be triggered programmatically or scheduled for the next boot.

flaviostutz:
The modes are the same: STANDALONE, CONFIG AND NORMAL

 - The application can put Homie in any one of the modes using Homie.setBootMode(...)
 - The application can schedule Homie to boot once in a determined mode using Homie.setBootModeNextBoot(...)
 - Overall refactoring on BootMode, removal of unused code and small fixes

Why this was implemented:
 - In my case, I want to enter config mode automatically when I detect that Wifi or MQTT are not working well in the next time the user turns on my device. Using this patch, I can call setBootModeNextBoot(...)
 - During Config mode, when there is a valid previous configuration, I will reboot the device automatically back to NORMAL mode after a timeout period (3min and so)
 - In my case, I don't have any buttons on the device, that's why I have to create "smart" ways to enter and exit config mode
 - All those new/refactored functions were teste by the App.

If you accept this PR, I will update the documentation. What do you think?

### DIFF
--- a/src/Homie.hpp
+++ b/src/Homie.hpp
@@ -37,10 +37,10 @@ class HomieClass {
   void __setFirmware(const char* name, const char* version);
   void __setBrand(const char* brand) const;
 
-  const BootMode MODE_UNDEFINED = BootMode::UNDEFINED;
-  const BootMode MODE_STANDALONE = BootMode::STANDALONE;
-  const BootMode MODE_CONFIG = BootMode::CONFIG;
-  const BootMode MODE_NORMAL = BootMode::NORMAL;
+  const static BootMode MODE_UNDEFINED = BootMode::UNDEFINED;
+  const static BootMode MODE_STANDALONE = BootMode::STANDALONE;
+  const static BootMode MODE_CONFIG = BootMode::CONFIG;
+  const static BootMode MODE_NORMAL = BootMode::NORMAL;
 
   HomieClass& disableLogging();
   HomieClass& setLoggingPrinter(Print* printer);
--- a/src/Homie.hpp
+++ b/src/Homie.hpp
@@ -37,10 +37,10 @@ class HomieClass {
   void __setFirmware(const char* name, const char* version);
   void __setBrand(const char* brand) const;
 
-  static const BootMode MODE_UNDEFINED = BootMode::UNDEFINED;
-  static const BootMode MODE_STANDALONE = BootMode::STANDALONE;
-  static const BootMode MODE_CONFIG = BootMode::CONFIG;
-  static const BootMode MODE_NORMAL = BootMode::NORMAL;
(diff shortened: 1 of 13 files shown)


src/Homie/Config.cpp
```diff
@@ -262,8 +281,8 @@ bool Config::patch(const char* patch) {
     return true;
 }
 
-BootMode Config::getBootMode() const {
-  return _bootMode;
+bool Config::isValid() {
```
euphi: bool Config::isValid() **const**

src/Homie/Config.cpp
```diff
@@ -168,26 +167,46 @@ void Config::erase() {
   if (!_spiffsBegin()) { return; }
 
   SPIFFS.remove(CONFIG_FILE_PATH);
-  SPIFFS.remove(CONFIG_BYPASS_STANDALONE_FILE_PATH);
+  SPIFFS.remove(CONFIG_NEXT_BOOT_MODE_FILE_PATH);
 }
 
-void Config::bypassStandalone() {
+void Config::setBootModeOnNextBoot(BootMode bootMode) {
```
euphi: This does not change the `Config` instance, so it can be defined as `const`.

src/Homie/Config.cpp
```diff
@@ -168,26 +167,46 @@ void Config::erase() {
...
+    bootModeFile.close();
+    Interface::get().getLogger().printf("Setting next boot mode to %d\n", bootMode);
+  }
 }
 
-bool Config::canBypassStandalone() {
-  if (!_spiffsBegin()) { return false; }
-
-  return SPIFFS.exists(CONFIG_BYPASS_STANDALONE_FILE_PATH);
+BootMode Config::getBootModeOnNextBoot() {
```
euphi: This method should be `const`.

flaviostutz:
@marvinroger, could you please see if this is OK? I have another pull request to be submitted, but I want to know if this is OK...

src/Homie.hpp
```diff
@@ -37,6 +37,11 @@ class HomieClass {
   void __setFirmware(const char* name, const char* version);
   void __setBrand(const char* brand) const;
 
+  static const BootMode MODE_UNDEFINED = BootMode::UNDEFINED;
```
marvinroger: For it to be consistent, can you please make the `BootMode` a `HomieBootMode` and include it the same way it works for https://github.com/marvinroger/homie-esp8266/blob/develop/src/HomieEvent.hpp? This way, the user could use `HomieBootMode::NORMAL` from his code.

flaviostutz:
@marvinroger, Please check if this is ok now.

flaviostutz:
I could not get my new commits to be included here. Closing and opening another pull request.

Issue #243, opened 2016-11-26
Bug when using getMqttClient() feature.


sheda:
Hi,
When using the `Homie.getMqttClient();` feature the OnMessage callback is called but topic argument is not complete.

This is due to the BootNormal.cpp OnMessage callback that suppress '/' and remplace those by '\0' vithin the topic buffer variable(which will be then given to our custom OnMessage callback).
Topic is shorten until node, property and  set field are behind the end_of_string.

I agree that this bug appears depending on in which order OnMessage callback are registered, but this seems a bit dirty when processing message to not check device-id while subscription are possible for any MQTT topic.

I made a fix to BootNormal.cpp, the fix consist in verifying that arriving message are destinated to current device-id (after processing for Broadcast messages)

Before submitting your issue, make sure:

- [ ] You're using a [stable release](https://github.com/marvinroger/homie-esp8266/releases), not the Git development version which is, by definition, unstable
- [X] You've read the documentation for *your* release, which is in the `docs/` folder of the `.zip` of the release you're using, especially the **Getting started** and **Troubleshooting** pages, which contain respectively the minimum required version of the dependencies, and some answsers to the most common problems
- [X] You're using the examples bundled in *your* release, which are in the `examples/` folder of the `.zip` of the release you're using. Examples in the latest git revision might not be backward-compatible with your release

Thanks!

marvinroger:
Thanks!

PR #244, opened 2016-11-26
[BootNormal] Fix topic message being shrunk when calling custom onMes…

sheda:
[BootNormal] Fix topic message being shrunk when calling custom onMessage callback

When using getMqttClient feature according to order onMessage callback are called
the homie one shring topic buffer (by adding '\0') and so following custom callback
doesn't get full topic

Related to Issue #243


marvinroger:
Did not think about this. Thanks! :+1:

Issue #250, opened 2016-12-02
IteadSonoffButton


Gulaschcowboy:
Hi,

after latest commits (don't ask me which one) the IteadSonoffButton example doesn't work anymore.
The ESP registers for `homie/<devicename>/+/+/set`, but it doesn't react to a publish on `homie/<devicename>/switch/on/set`

Any hints?
Thanks
Alex

elbowz:
I have found the same bug...this [commit](https://github.com/marvinroger/homie-esp8266/commit/14994b3225e51fa8a4c8430153fca76716f14729) is the problem!

This because ```device_topic``` and ```deviceId``` is not the same :)

devices/**device ID**/node-ID/property
instead the ```device_topic``` is: *node-ID/property*

```
   for (uint16_t i = 0; i < strlen(device_topic); i++) {
     if ((device_topic[i] != messageDeviceId[i]) || (device_topic[i] == '/' && messageDeviceId[i] != '\0')) {
         Interface::get().getLogger() << F("DEBUG: USCITA!!! ") << device_topic[i] << messageDeviceId[i]  << endl;
       return;
     }
   }
```

Gulaschcowboy:
@elbowz : Yes ,that's the issue. Thanks

sheda:
Hi Guys,
My bad, this issue is on me. @elbowz You are right I don't use the right topic to check the device (got confused by the name, and circumstances of my sketch made this still work for me :-/ ).
Please find fix [here](https://github.com/sheda/homie-esp8266/commit/7d6e20753c6fcdbf076d9001fd848a7d6f930c7e)
Apologies for any inconveniences.

elbowz:
Thanks for your rapidly solution! :)

marvinroger:
Merged, thanks guys!

PR #253, opened 2016-12-03
[BootNormal] Fix, Bad topic used for device name detection

sheda:
@marvinroger Meaculpa, previous commit was completly wrong. This one should fix it.
Please see issue #250 for more details

More generally, with the direct access to MQTT client(getMqttClient()) and so topic arriving could be anything, don't you think "brand/device-id" should be checked each time before processing any topic?
Checking only from device assumes that brand is the same. But those can be different at the content and the length level implying shift in the device-id comparison. And Maybe creating false positive if brand name contain device-id name...

Update: rebased fix for PR


marvinroger:
Don't worry, I should have spotted the bug being merging, too... Anyway, thanks!

PR #274, opened 2017-01-03
CORS for /wifi/connect

keyeh:
CORS for `/wifi/connect`

Using the same code for handling CORS requests for `/config`.

Use case: if the configurator is a separate website/webapp.


marvinroger:
Thanks! 😃

PR #280, opened 2017-01-06
Log wifi client IP address upon connection.

maruel:
This removes the need to hunt the device IP when debugging connectivity issues.


marvinroger:
Thanks!

Issue #273, opened 2017-01-03
Defining nodes at runtime


snikch:
When I define a node outside the `setup` function, it works fine, however defining it first thing inside the `setup` function, it causes a reset after the wifi is connected. I assume because of this, defining nodes at runtime is not a supported feature.

```cpp
void setup() {
  HomieNode n1("switch1", "switch");
  Serial.begin(115200);
  Homie_setFirmware(FW_NAME, FW_VERSION);
  Homie_setBrand(BRAND_NAME);
  Homie.setup();
  Serial.println("Setup complete");
}

void loop() {
  Homie.loop();
}
```

Causes

![image](https://cloud.githubusercontent.com/assets/115958/21600477/cffb7528-d1e1-11e6-96cc-0ee632601a20.png)

However this works fine

```cpp

HomieNode n1("switch1", "switch");
void setup() {
  Serial.begin(115200);
  Homie_setFirmware(FW_NAME, FW_VERSION);
  Homie_setBrand(BRAND_NAME);
  Homie.setup();
  Serial.println("Setup complete");
}

void loop() {
  Homie.loop();
}
```

I'd like to be able to implement a class that can be configured to add multiple, or just single nodes, at runtime via the configuration.json. Is this completely whack and I should avoid this totally, or is there something simple I'm just missing? (This seems to be the simplest test case I can reduce my scenario down to fyi, not quite the code that I'm attempting to run).

egon0:
ha, i thought about the same use-case, home-nodes configurable via config.json. hope we get the possibility sometime. at the moment i have a homie-base-sketch and some git branches for the many different sensor-combinations i use. would love to have one fw for all and activate/configure the nodes via config.json

gtom:
It seems, my problem is very similar to this item: Before I declare my HomieNode in my sketch, I want to create a node-name based on MAC-address. But the earliest time to do so is in setup(). But here it is too late for a HomieNode-Statement. 

Does someone know, how to do this as a global variable as stated above by euphi? It doesn't work with just
`HomieNode myTestNode;`

My reason for this is the same as for egon0 two postings above: I want one firmware for a bunch of devices. Each device-behavior is defined in this firmware and is selected by MAC-address.

euphi:
Dynamic **creation** of HomieNode objects should work with the current version of Homie. 
I haven't tested it, but if you create them with `new`, you can do this in setup() and it should work fine.
[...]

PR #286, opened 2017-01-09
Added destructor to HomieNode, that will abort()

euphi:
To "fix" https://github.com/marvinroger/homie-esp8266/issues/273, I added a destructor to HomieNode that will abort after sending an error message to the Logger.

This should help C++ novices to avoid common pitfals as defining a Node on the stack (e.g. in setup())


euphi:
Sorry, I was not able to "squash" the commits and there "merge" back before, so the only real commit is https://github.com/marvinroger/homie-esp8266/pull/286/commits/134844ddfc64e4d0235d3d77a86a3582e516063f . However, the only commited change is only this commit. All the other commits are already in line with your develop branch.

 If somebody explains me, how to squash the merge commits before, I will do it in the future :-).

marvinroger:
Awesome, thank you. I have no idea for the Git problem, though, but I've squashed the commit directly on GitHub so everything's fine. 😃

PR #290, opened 2017-02-06
Fixing loopFunction being called too early before MQTT connection is established

furyfire:
Previously the check was 
` if (!Interface::get().connected) { return; }`

Fixed brackets to ensure loopFunction is only called when homie is connected


marvinroger:
Thanks a lot @furyfire 👍

PR #292, opened 2017-02-07
Length checks on input fields. 

furyfire:
Added length checks on some strings.
Fixed Compiler warning - "list-initializer for non-class type must not be parenthesized"


marvinroger:
Thanks!

furyfire:
And yeah sorry about the warnings. I better start using cpplint so I can check the code properly before making a pull request.

marvinroger:
Don't worry, it's fixed upstream.

PR #296, opened 2017-02-16
bug: DeviceId returns incomplete MAC address

bertmelis:
snprintf works with n-1 characters

expected behaviour: DeviceId is 12 characters
behaviour: only 11 characters are returned
(resulting in a "incomplete" mac address)

DeviceID.hpp also states
static char _deviceId[MAX_MAC_STRING_LENGTH + 1];


marvinroger:
Thanks!

Issue #299, opened 2017-02-18
after just 50 days uptime jumps to 40 years


cpan-mila:
after approximately 50 days the number of milliseconds returned from the millis() function overflows. this is not properly handled in Uptime::update (src/Homie/Uptime.cpp). this leads to "uptimes" of over 40 years, and that after just 50 days.

bertmelis:
Would this code be of any help?
https://gist.github.com/AndiSHFR/1f8b71dbeae6ee9073e0

NemoN:
+1

i would be interessted in a solution too.

bertmelis:
I'll try to implement the code above, but how should I test?
I don't want to wait for 50 days before I know it's working.
(Besides that, I'm too busy playing with the devices that I'll never make it up to 50 days.)

marvinroger:
@bertmelis just make a dummy `millis()`-like function that returns 4,294,967,295 on the first call and 10000 on the second call. After the second call, the number of elapsed seconds should be about 4,294,977... And not 40 years!

marvinroger:
Fixed by @bertmelis, thanks!

PR #302, opened 2017-02-23
#299 Handling millis() overflow in uptime

bertmelis:
Corrected type declaration for correct overflow handling of millis
function in uptime calculation

### DIFF
--- a/src/Homie/Uptime.cpp
+++ b/src/Homie/Uptime.cpp
@@ -8,7 +8,7 @@ Uptime::Uptime()
 }
 
 void Uptime::update() {
-  uint64_t now = millis();
+  uint32_t now = millis();
   _seconds += (now - _lastTick) / 1000UL;
   _lastTick = now;
 }
--- a/src/Homie/Uptime.hpp
+++ b/src/Homie/Uptime.hpp
@@ -11,6 +11,6 @@ class Uptime {
 
  private:
   uint64_t _seconds;
-  uint64_t _lastTick;
+  uint32_t _lastTick;
 };
 }  // namespace HomieInternals


marvinroger:
What's the explanation behind the fact that it fixes the issue?

marvinroger:
Oh, maybe because millis returns an `unsigned long` which is, on arduino, 32 bits and not 64?

Nzbuu:
I suspect it's due to the wraparound handling in `(now - _lastTick)`

Nzbuu:
Would it be better to use `unsigned long` rather than `uint32_t` or `uint64_t`?

bertmelis:
millis returnes an unsigned long indeed, which is uint32_t in normal C++.
So the rollover didn't work properly because not everything rolled over (or how should I explain).

Since the rest of the Homie code follows C++ variable declaration uses the standard C++ names, I also used these instead of the Arduino names like "unsigned long". It's better to stick to the general C++ names wrt portability. (as I've read somewhere.)

bertmelis:
If memory is an issue, you could also make the `uint64_t  seconds` and `uint32_t`.
Hence, it would rollover after +130 years, which is quite a while for my little IoT-devices.

bertmelis:
I did a quick scan of the rest of the Homie code. Everywhere else, `uint32_t` variables are used in conjuction with the `millis()` function.

marvinroger:
Yes, good catch! Thanks a lot for your time.

bertmelis:
As @euphi pointed out, the uptime could be not that accurate. Could someone review this solution:

```C++
void Uptime::update() {
  uint32_t now = millis();
  _milliseconds += (now - _lastTick);
  _lastTick = now;
}
uint64_t Uptime::getSeconds() const {
  return _milliseconds / 1000ULL;
}
```

So instead of storing seconds, milliseconds are stored. getSeconds obvoisly still return the uptime in seconds. The cost of this is that the uptime get rolled over earlier. But the max uptime is still 18446744073709551615 milliseconds.

PR #304, opened 2017-02-23
Fix for #299

furyfire:
This should fix bug #299 #302
I do not believe #302 fixes the issues correctly

### DIFF
--- a/src/Homie/Uptime.cpp
+++ b/src/Homie/Uptime.cpp
@@ -14,6 +14,7 @@ void Uptime::update() {
   if (now < _lastTick) {
     _rolloverCount++;
   }
+  
   _lastTick = now;
   _seconds =  (0xFFFFFFFF / 1000 ) * _rolloverCount + (_lastTick / 1000);
 }
--- a/src/Homie/Uptime.cpp
+++ b/src/Homie/Uptime.cpp
@@ -4,15 +4,20 @@ using namespace HomieInternals;
 
 Uptime::Uptime()
 : _seconds(0)
-, _lastTick(0) {
+, _lastTick(0)
+, _rolloverCount(0)   {
 }
 
 void Uptime::update() {
-  uint64_t now = millis();
-  _seconds += (now - _lastTick) / 1000UL;
+  uint32_t now = millis();
+  
+  if (now < _lastTick) {
+    _rolloverCount++;
+  }
   _lastTick = now;
+  _seconds =  (0xFFFFFFFF / 1000 ) * _rolloverCount + (_lastTick / 1000);
 }
 
-uint64_t Uptime::getSeconds() const {
+uint32_t Uptime::getSeconds() const {
   return _seconds;
 }
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,7 +3,7 @@ If you have some questions or if you need help, some people might help you on th
 
 Before submitting your issue, make sure:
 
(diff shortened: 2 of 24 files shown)


bertmelis:
can you explain what's not working in my fix for the uptime update?

euphi:
@bertmelis : Just a guess from me: It fixes the rounding issue due to addings seconds only.
The old code from @marvinroger that you have fixed, always cuts the remainding milliseconds.
The new codes uses the actual uptime in milliseconds plus keeps track of the amount of overflows, so it is much more accurate.

furyfire:
Ehh now I'm no longer sure. I'll have to examine that a little deeper.

bertmelis:
I'm not sure either. @euphi is right about the accuracy. By dividing the millis by 1000 in the original solution, the remainder gets thrown away. So it could be that the uptime drifts (accummulatively) by a seconds every time the uptime gets updated.

A solution would be to not to store the total number of seconds as the uptime, but the total number of milliseconds. The function `getSeconds` can return the` _milliseconds / 1000` however for not to break compatability.

amayii0:
Just curious, what are the use cases where you want to use millis() let's say on periods longer than a few days/weeks?
For simple fast loops, using a few days (inc ms) should be fine and accurate.
For very long loops, using a bigger scale like seconds or even minutes/hours should be precise enough.
Am I missing something here?
Also what about exposing individuals days/hours/minutes/ms parts?

src/Homie/Uptime.cpp
```diff
@@ -4,15 +4,21 @@ using namespace HomieInternals;
...
 void Uptime::update() {
   uint32_t now = millis();
-  _seconds += (now - _lastTick) / 1000UL;
+  
+  if (now < _lastTick) {
+    _rolloverCount++;
+  }
+  
   _lastTick = now;
+  _seconds =  (0xFFFFFFFF / 1000 ) * _rolloverCount + (_lastTick / 1000);
```
Nzbuu: I'm not sure this is right. I think that you loose 1 milli-second every roll over:

If the limit is 0xFF, the previous tick is 0x01 and the current tick is 0x00, then this mean that 0x100 ticks (seconds = 0x100 ./ tickLSB) have happened. However, this logic would give rolloverCount = 1, lastTick = 0, and seconds = 0xFF / tickLSB.

Nzbuu: Actually, it might be worse than that. `0xFFFFFFFF / 1000` is integer division, giving the result `4294967`, which is missing the fractional part of 0.295. So it loses 295 milliseconds each rollover (actually 296, as above).

It's OK to round down to the nearest integer number of seconds overall since that's the objective, but I think it's probably not OK to be out by more than 1 second.

Nzbuu: How about:

```
_seconds = (((uint64_t) rollover_count << 32) + last_tick) / 1000
```

I would store the uint64_t value of `((uint64_t) rollover_count << 32)` rather than the uint32_t `rollover_count`, since I think it needs less memory overall.

bertmelis:
```C++
void Uptime::update() {
  uint32_t now = millis();
  _milliseconds += (now - _lastTick);
  _lastTick = now;
}

uint64_t Uptime::getSeconds() const {
  return (_milliseconds / 1000ULL);
}
```

every update the elapsed time in milliseconds is added and the remainder of the division is only thrown away while converting to seconds, but is kept for future updates..

Should this do the trick? `_milliseconds` would be a `uint64_t` also.

bertmelis:
@amayii0 I agree on the fact that the uptime in milliseconds is way too precise. But you have to keep track of time somehow. You could also keep the uptime in seconds or even minutes. But how are you going to update this value? You eventually end up using timer interrupts or using millis().

Issue #301, opened 2017-02-21
keepAlive value hardcoded


bertmelis:
EDIT: the opening post is dated...

I've been using the Ticker.h library for my timer based events. This library generates a timer interrupt and executes a routine.

However, Homie and Ticker seems a no good marriage, especially if you've got a timer at 60 secods.
Almost every time the ticker issues an interrupt, my MQTT connection is lost.

I've tested the same sketch but replaced the ticker-interrupt by the common "millis()-lastmillis"-construction and no more disconnects were seen.

I know about blocking code, but my ISR is only flagging a boolean variable so is fast enough.

Or is there something wrong with the interrupt system itself? (troubles with pointers...?)

Any tips for testing and debugging?

bertmelis:
After a long evening of hair pulling, I found out the keepalive value for the mqttclient is hardcoded in the setup function of the normal boot process.
So even if we set it to a different value, it get's overwritten.

I see no point in having it hardcoded as the async mqtt lib already defaults to 15sec if no value is given.

OK to just delete?

(The problem in my sketch was not related to any incompatabilities. It just disconnected because of poor wifi in my basement and a too short keepalive.)

PR #305, opened 2017-02-24
Improve uptime accuracy

bertmelis:
Improve accuracy of the uptime calculation by storing uptime as milliseconds.
Conversion to seconds is only done when preparing to main code.

Drawback: the maximum uptime is a lot lower than before, but still way more than I'd ever live.

### DIFF
--- a/src/Homie/Boot/BootConfig.cpp
+++ b/src/Homie/Boot/BootConfig.cpp
@@ -42,7 +42,7 @@ void BootConfig::setup() {
     WiFi.softAP(apName);
   }
 
-  snprintf(_apIpStr, MAX_IP_STRING_LENGTH, "%d.%d.%d.%d", ACCESS_POINT_IP[0], ACCESS_POINT_IP[1], ACCESS_POINT_IP[2], ACCESS_POINT_IP[3]);
+  snprintf(_apIpStr, MAX_IP_STRING_LENGTH + 1, "%d.%d.%d.%d", ACCESS_POINT_IP[0], ACCESS_POINT_IP[1], ACCESS_POINT_IP[2], ACCESS_POINT_IP[3]);
 
   Interface::get().getLogger() << F("AP started as ") << apName << F(" with IP ") << _apIpStr << endl;
   _dns.setTTL(30);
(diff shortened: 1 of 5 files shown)


bertmelis:
PS I also committed the keepalive to my fork so it showed up here. I'm in the assumption that the hardcoded keepalive value for MQTT can be deleted as mentioned in #301.

bertmelis:
I didn't test the config mode but you're probably right as this isn't a "normal" string but a variable "IPaddress". I was a bit hasty/blunt.

PS I'm going crazy of my autocorrect. It's always correcting to a like Dutch word.

bertmelis:
I just tested online using  idone.com:

```C++
#include <iostream>
#include <cstring>
using namespace std; // consider removing this line in serious projects

int main() {
	int MAX_IP_STRING_LENGTH = strlen("192.168.129.100");
	char _apIpStr[MAX_IP_STRING_LENGTH];
	snprintf(_apIpStr, MAX_IP_STRING_LENGTH + 1, "%d.%d.%d.%d", 192, 168, 123, 123);
	cout << _apIpStr << "\n";
	return 0;
}
```
without the `+1`, the output is `192.168.123.12`.

By the way: it has nothing to do with the actual IP address you are using. It's the output of this IP address in the logger and in MQTT that's incorrect.

avdleeuw:
Bert, 

good point. But unfortunately, MAX_IP_STRING_LENGTH is not determined with a 'strlen' function. 

In Limits.hpp, you can find this definition (notice the 'sizeof'): 

const uint8_t MAX_IP_STRING_LENGTH = sizeof("123.123.123.123");

When I add this to your sample code and run it on ideone.com as well, it becomes obvious where the difference stems from. 

```
#include <iostream>
#include <cstring>
using namespace std; // consider removing this line in serious projects

int main() {
	int MAX_IP_STRING_LENGTH = strlen("192.168.129.100");
	char _apIpStr[MAX_IP_STRING_LENGTH];
	snprintf(_apIpStr, MAX_IP_STRING_LENGTH + 1, "%d.%d.%d.%d", 192, 168, 123, 123);
	cout << _apIpStr << "\n";
	cout << "Strlen:" << MAX_IP_STRING_LENGTH << endl;
	cout << "Sizeof:" << sizeof("192.168.129.100") << endl;
	return 0;
}
```

Output: 

```
192.168.123.123
Strlen:15
Sizeof:16
```

bertmelis:
You're totally right.

PR #314, opened 2017-03-06
Fix #301 remove hardcoded keepalive

### LINKED ISSUE #301: keepAlive value hardcoded

bertmelis:
Fix for #301: Remove hardcoded KeepAlive for MQTT connection.
Default value of 15sec is already present in AsyncMQTT library.

Custom keepalive value got overwritten by the hardcoded one, but Homie would ping based on the custom setting.


marvinroger:
Thanks!

PR #315, opened 2017-03-06
Improve uptime accuracy

bertmelis:
improve uptime accuracy by storing milliseconds and only rounding when publishing value
drawback: less time before rollover uptime, but still long enough


marvinroger:
Thanks. 😄

PR #335, opened 2017-05-11
 IHomieSetting::settings first, HomieNode-settings second!

euphi:
Set initialization priority of IHomieSetting::settings to value of highest allowed priority (101).

This allows other static variables to be of type HomieSetting< T>.

See https://github.com/euphi/HomieNodeCollection/blob/master/src/RGBWNode.cpp for example.

Note: It makes sense to have a static HomieSetting member, if you have a class that may be instantiated multiple times. (e.g. two LED Strips connected to the same ESP8266, as shown in https://github.com/euphi/ESP-LEDCtrl).

### DIFF
--- a/src/HomieSetting.cpp
+++ b/src/HomieSetting.cpp
@@ -2,7 +2,7 @@
 
 using namespace HomieInternals;
 
-std::vector<IHomieSetting*>  __attribute__ ((init_priority (101))) IHomieSetting::settings;
+std::vector<IHomieSetting*> __attribute__((init_priority (101))) IHomieSetting::settings;
 
 template <class T>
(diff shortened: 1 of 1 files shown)


euphi:
Additional remark:

The C++ standard gives no information about initialization order of static object in different compilation units.

The attribute `init_priority` is a GCC extension that allows the linker to sort the objects; 101 ist the lowest value (highest priority) that is available for normal use.

In my case it seems that my static `HomieSetting<long> RGBWNode::fadeDelay ("RGBfadeDelay", "fade delay"); ` was initialized before ` IHomieSetting::settings` . Interestingly the call to `push_back` in the constructor of `HomieSetting<long>` just did nothing?! (I expected a crash, but the vector just stayed empty).

PR #360, opened 2017-06-23
Breakup load into separate functions

jamesgol:
This splits off the actual loading of the JSON file and parsing into a new function. The idea is that eventually I'd like to make modifying the configuration file available to the code base.

I'm not sure what the best method for exposing the configuration is, but this is a start.

### DIFF
--- a/src/Homie/Config.cpp
+++ b/src/Homie/Config.cpp
@@ -17,27 +17,25 @@ bool Config::_spiffsBegin() {
   return _spiffsBegan;
 }
 
-bool Config::load() {
-  if (!_spiffsBegin()) { return false; }
-
-  _valid = false;
+JsonObject& Config::loadJSONObject() {
+  if (!_spiffsBegin()) { return JsonObject::invalid(); }
 
   if (!SPIFFS.exists(CONFIG_FILE_PATH)) {
     Interface::get().getLogger() << F("✖ ") << CONFIG_FILE_PATH << F(" doesn't exist") << endl;
-    return false;
+    return JsonObject::invalid();
   }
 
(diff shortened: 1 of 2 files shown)


src/Homie/Config.cpp
```diff
@@ -49,6 +47,17 @@ bool Config::load() {
   JsonObject& parsedJson = jsonBuffer.parseObject(buf);
```
euphi: `jsonBuffer` is a local variable, so  you create a reference to something internal to this local variable.
The local variable gets destroyed when leaving `Config::load()`, this (hopefully ;-) frees the buffer of `parsedJson`.

src/Homie/Config.cpp
```diff
@@ -49,6 +47,17 @@ bool Config::load() {
...
+    return JsonObject::invalid();
+  }
+  return parsedJson;
+}
+
+
+bool Config::load() {
+  _valid = false;
+  JsonObject& parsedJson = Config::loadJSONObject();
+  if (!parsedJson.success()) {
```
euphi: `parsedJson` is freed memory, so behaviour is undefined. (It will work in most cases anyhow, because no one else has overwritten the memory region yet... but technically this is just good luck....)

jamesgol: It's been a decade or two since I've done any real C++ coding.  Thanks for the sanity check.

timpur:
This has been done in #458 differently and in a safer manner (event though my c++ is basic ... lots of googling) , thus im closing, if you have any comments please add :)

jamesgol:
Excellent.  I'll have to make some time to test it out.

PR #375, opened 2017-07-27
Change topic check for OTA upload

jesserockz:
Fixes the module crashing as soon as firmware is published into the topic


marvinroger:
How does adding an `s` to "firmware" fix the crash? A singular `firmware` makes more sense anyway.

jesserockz:
The topic being subscribed to is /firmwares/*
It looks like one was changed but not the other.

https://github.com/marvinroger/homie-esp8266/blob/develop/src/Homie/Boot/BootNormal.cpp#L299

marvinroger:
Oh, right. Would you mind do the opposite, rename both to `firmware` in singular?

marvinroger:
Awesome, thanks a lot!

PR #378, opened 2017-08-03
Fix: not returning a value in a method that returns non-void

euphi:
`HomieClass& HomieClass::setConfigurationApPassword(const char* password)` did not returned a reference to the Homie instance. This PR fixes this.


marvinroger:
Thanks! :+1:

PR #379, opened 2017-08-03
Fix crash when starting up without any defined node

jbliesener:
Skip node publication if HomieNode::nodes.size() == 0

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -404,7 +404,7 @@ void BootNormal::_onMqttMessage(char* topic, char* payload, AsyncMqttClientMessa
     && strcmp(_mqttTopicLevels.get()[0], Interface::get().getConfig().get().deviceId) == 0
     && strcmp_P(_mqttTopicLevels.get()[1], PSTR("$implementation")) == 0
     && strcmp_P(_mqttTopicLevels.get()[2], PSTR("ota")) == 0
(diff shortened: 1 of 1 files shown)


marvinroger:
Thanks! :wink:

luebbe:
Why is it "firmwares" in plural? I would expect that only one firmware can run on a specific device at any time.

marvinroger:
Oh, I miss this one, nice catch @luebbe 
@jbliesener how does adding an `s` to `firmware` fix the crash?

jbliesener:
Well, actually the second fix was for my repository only, as it replicates another pull request (#375). I applied it to my code, as I had problems with OTA upload. I wasn't even aware that it would automatically be replicated to this pull request (I should have opened a new branch to avoid this).

Anyway, the additional "s" makes the OTA update run through. Although I have two observations that I don't know if they're related:

1. OTA update runs through and starts the new firmware, but there seems to be another crash/stack dump just before (or after) rebooting. I will take a look at this tonight.
2. Maybe, line 752 of BootNormal also needs to be adjusted. The longest topic may now be "/$implementation/ota/firmwares/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx". So, the additional "s" makes the topic one byte longer and we have a slash and 32 bytes of the MD5 sum string. 

Also, in order to be able to get line numbers in the stack trace, I added a "build_flags = -g" line to platform.ini. Helps a lot...

PR #380, opened 2017-08-04
Prevent Wifi reconnect before reboot

jbliesener:
When Homie schedules a reboot, it disconnects Wifi, but tries to reconnect afterwards, which results in a crash. 

This seems to happen because `ESP.restart()` - contrary to what one would expect - returns and leaves the program running for a short while, before actually restarting the processor. 

This patch disables the interface, so that it won't reconnect anymore after the reboot has been requested. A different way to achieve the same result may be a `while(1);` after `ESP.restart()`.

### DIFF
--- a/src/Homie/Boot/BootConfig.cpp
+++ b/src/Homie/Boot/BootConfig.cpp
@@ -385,6 +385,7 @@ void BootConfig::loop() {
   if (_flaggedForReboot) {
     if (millis() - _flaggedForRebootAt >= 3000UL) {
       Interface::get().getLogger() << F("↻ Rebooting into normal mode...") << endl;
+      Interface::get().flaggedForSleep = true;
       Serial.flush();
       ESP.restart();
     }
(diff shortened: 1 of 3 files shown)


marvinroger:
Good catch! However, I'd prefer not to workaround this issue by setting the sleep flag. Can you test the while (true) approach?

jbliesener:
Nope, doesn't work. `BootNormal::_onWifiDisconnected` is called **asynchronously** from the ESP libs (even during the `while(1);`) and in turn calls `BootNormal::_wifiConnect`, which tries to reconnect Wifi on an interface that was shut down.

You also could consider `BootNormal::_flaggedForReset` in `BootNormal::_wifiConnect` and change the condition to 

```
if (!Interface::get().flaggedForSleep && !_flaggedForReset) {
```

However, I wouldn't say that using `InterfaceData::flaggedForSleep` is a workaround. Maybe you could just rename the flag to `doNotInit`.

marvinroger:
Okay, cleanest way: make both WiFiEventHandler `= nullptr` when restarting, therefore the WiFi handler won't be called. :wink:

timpur:
I belives this was fixed by #432

PR #393, opened 2017-09-03
🤔 Add details on how to interact with range property

LiyouZhou:
Add hint to help people figure out how to interact with
range properties. Especially document the `_` separator.

PR #396, opened 2017-09-07
Document the topic name used for range properties

lorenwest:
I couldn't find an example topic name for range properties in the documentation, and had to search through issue discussions to find it. 

Thought the next person would appreciate an example in this file.


LiyouZhou:
I have a pr in flight https://github.com/marvinroger/homie-esp8266/pull/393 good to know I'm not the only one finding this irritating.

lorenwest:
Deleting as duplicate of #393

Issue #397, opened 2017-09-08
Python OTA updater script: Wrong datatype for firmware buffer


euphi:
The new OTA updater script at https://github.com/marvinroger/homie-esp8266/tree/develop/scripts/ota_updater does not work for me, because the datatype of the firmware buffer is 'str', which results in an error in paho-mqtt library, because it trys to convert it to unicode ?!

I dont really understand why this happens, especially because the firmware file is opened in 'rb' mode, which should results in reading the file as binary_array(?!)

However, the type of `firmware` is `str`, see error log (I added a line to print `firmware.__class__` to see the datatype):

```
Connecting to mqtt broker cubietruck on port 1883
Connected with result code 0
Waiting for device info...
Publishing new firmware with checksum b09270aaa47fc5461ae9e16504a3942f
homie/balkctrl2/$implementation/ota/firmware/b09270aaa47fc5461ae9e16504a3942f
<type 'str'>
Traceback (most recent call last):
  File "./ota_updater.py", line 154, in <module>
    args.broker_password, args.base_topic, args.device_id, firmware)
  File "./ota_updater.py", line 110, in main
    client.loop_forever()
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1378, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 897, in loop
    rc = self.loop_read(max_packets)
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1177, in loop_read
    rc = self._packet_read()
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 1766, in _packet_read
    rc = self._packet_handle()
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2240, in _packet_handle
    return self._handle_publish()
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2415, in _handle_publish
    self._handle_on_message(message)
  File "/home/ian/.local/lib/python2.7/site-packages/paho/mqtt/client.py", line 2575, in _handle_on_message
    self.on_message(self, self._userdata, message)
  File "./ota_updater.py", line 85, in on_message
    client.publish(topic, userdata['firmware'])
[...]

PR #403, opened 2017-09-24
Interpret firmware file as an bytearray

LiyouZhou:
Fix https://github.com/marvinroger/homie-esp8266/issues/397


LiyouZhou:
@marvinroger please take a look.

marvinroger:
Thanks. :+1:

Issue #178, opened 2016-10-15
2 Bugs? Config and $online after todays commit


Gulaschcowboy:
Hi Marvin,

first off all: Great project, I love it.

Two things I discovered after I did a git pull today.

1. to devices/<id>/$online alsways false is submitted
so I thought, ok, let's do a reset. But I discovered another thing:

2. Config always reports: "There was an error while sending the configuration: Invalid or too big JSON
Please retry or report if it looks like a bug. "
I used latest app version and also the online app of you.

Thanks,
Alex

Gulaschcowboy:
Update:
I tried curl method and found this behaviour:

Sending this config it works:
{
  "name": "bee2",
    "wifi": {
    "ssid": "my-ssid",
    "password": "my-pass"
     },
     "mqtt": {
     "host": "omd.atze.lan",
     "port": 1883
     },
     "ota": {
     "enabled": true
      }
}

If I add auth credentials like:
{
  "name": "bee2",
    "wifi": {
    "ssid": "my-ssid",
    "password": "my-pass"
     },
     "mqtt": {
     "host": "omd.atze.lan",
     "port": 1883
     "auth": true,
     "username": "my-mqttuser",
     "password": "my-mqttpass"
     },
     "ota": {
     "enabled": true
      }
}

I get the response: {"success":false,"error":"Invalid or too big JSON"}

marvinroger:
I know what's wrong with the second issue, it's because ArduinoJson tries to change the content of the internal buffer of an Arduino String, which is wrong. That's not actually an issue from ArduinoJson, though.

What do you mean by `two devices//$online always false are submitted`? Like, at startup? Is an `$online true` submitted or not at all?

Gulaschcowboy:
Hi Marvin,

after recompiling and flashing newest github version the device always posts "false" to devices/<dev-id>/$online

What is solution or workaround to issue 1?

marvinroger:
I did not fix anything yet, I'll tell you when it's done.

The device never post `false` to `$online`, but the broker implicitely sends this when the device is disconnected from the TCP socker. Or, you use `prepareToSleep()`.

Can you post your sketch?

Gulaschcowboy:
[homie-Bienen1.txt](https://github.com/marvinroger/homie-esp8266/files/531508/homie-Bienen1.txt)

Sorry, couldn't post inline, as github api eats some code parts

marvinroger:
[...]

PR #416, opened 2017-10-29
Backport fix for issue #178 into master/stable branch

nkaminski:
The example JSON configuration provided at http://marvinroger.github.io/homie-esp8266/stable/5.-JSON-configuration-file/ fails to load when using the master/"stable" branch of this library when built against the latest version of ArduinoJSON due to issue #178. Therefore, I would recommend incorporating this fix into the stable branch since such bug severely limits functionality, and the fix for such does not break compatibility with existing implementations.


marvinroger:
Thanks 👍

PR #425, opened 2017-11-15
AsyncWebServer + Refactoring

timpur:
Moved Config Mode to use AsyncWebServer + lots of refactoring.

Also plan on continuing to refactor homie.

### DIFF
--- a/library.json
+++ b/library.json
@@ -31,7 +31,10 @@
     {
       "name": "Bounce2",
       "version": "^2.1.0"
-    }
+    },
+	{
+	  "name": "ESPAsyncWebServer"
+	}
   ],
   "export": {
     "include": [
--- a/src/Homie/Boot/BootConfig.cpp
+++ b/src/Homie/Boot/BootConfig.cpp
@@ -58,18 +58,18 @@ void BootConfig::setup() {
   });
   _http.on("/device-info", HTTP_GET, [this](AsyncWebServerRequest *request) { _onDeviceInfoRequest(request); });
   _http.on("/networks", HTTP_GET, [this](AsyncWebServerRequest *request) { _onNetworksRequest(request); });
-  _http.on("/config", HTTP_PUT, [this](AsyncWebServerRequest *request) { _onConfigRequest(request); });
+  _http.on("/config", HTTP_PUT, [this](AsyncWebServerRequest *request) { _onConfigRequest(request); }).onBody(BootConfig::_parsePost);
   _http.on("/config", HTTP_OPTIONS, [this](AsyncWebServerRequest *request) {  // CORS
     Interface::get().getLogger() << F("Received CORS request for /config") << endl;
     _sendCROS(request);
   });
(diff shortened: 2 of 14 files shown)


library.json
```diff
@@ -31,7 +31,10 @@
     {
       "name": "Bounce2",
       "version": "^2.1.0"
-    }
+    },
+	{
+	  "name": "ESPAsyncWebServer"
```
marvinroger: Can you add a "version" field with the `^`, so that if there's a breaking change it still works.

src/Homie/Boot/BootConfig.cpp
```diff
@@ -278,99 +295,79 @@ void BootConfig::_onDeviceInfoRequest() {
...
-    Interface::get().getLogger() << F("✖ Device already configured") << endl;
-    String errorJson = String(FPSTR(PROGMEM_CONFIG_JSON_FAILURE_BEGINNING));
-    errorJson.concat(F("Device already configured\"}"));
-    _http.send(403, FPSTR(PROGMEM_CONFIG_APPLICATION_JSON), errorJson);
+    _SendJSONError(request, F("✖ Device already configured"), 403);
     return;
   }
 
-  StaticJsonBuffer<MAX_JSON_CONFIG_ARDUINOJSON_BUFFER_SIZE> parseJsonBuffer;
-  std::unique_ptr<char[]> bodyString = Helpers::cloneString(_http.arg("plain"));
```
marvinroger: What's wrong with `unique_ptr<>`?

timpur: TBH not the best a c++, dont know all its ins and outs, but [_tempObject](https://github.com/me-no-dev/ESPAsyncWebServer/blob/8604f67a5ff3d48ddc725a8606dcef774d9d5567/src/WebRequest.cpp#L91) is freed by the request when done. Thus is there need for this? Id love to know best practices.

library.json
timpur: Problem, they dont have a version [https://github.com/me-no-dev/ESPAsyncWebServer](https://github.com/me-no-dev/ESPAsyncWebServer). Shall i just add a commit hash ?

timpur:
@marvinroger im planing on doing more refactoring. Are you okay with this? Any thing you would like me to change in my coding style? Im trying to follow the same style as is, but as ive said, but the best at c++. So any tips are welcomed.

library.json
marvinroger: OK, nevermind then. 😉

src/Homie/Boot/BootConfig.cpp
marvinroger: No, indeed, you're right. Didn't know how ESPAsyncWebServer worked.

marvinroger:
@timpur of course I am okay, that's great for the project. As you probably noticed, I am unable to maintain the project currently, so definitely, go ahead! 👍

Issue #429, opened 2017-11-18
Unable to compile (Wifi plug firmware) 


Dwgystyl:
I'm working through http://www.hagensieker.com/blog/page/?post_id=44&title=esp8266-hack-of-inexpensive-wifi-outlet

and when it tries to compile i get the following. 

C:\Users\Username\Documents\Arduino\libraries\homie-esp8266-develop\src\Homie\Boot\BootConfig.cpp: In member function 'virtual void HomieInternals::BootConfig::setup()':
C:\Users\Username\Documents\Arduino\libraries\homie-esp8266-develop\src\Homie\Boot\BootConfig.cpp:48:39: error: passing 'const IPAddress' as 'this' argument of 'String IPAddress::toString()' discards qualifiers [-fpermissive]

   _apIpStr = ACCESS_POINT_IP.toString();

exit status 1
Error compiling for board Generic ESP8266 Module.

I can track down the exact line it is calling too, but I am not familiar enough to figure out what or why the problem is.

timpur:
Thats interesting have what version of esp8266 Arduino core are you using ?

ngalfas:
I have exaxtly the same error.    I have Arduino IDE 1.8.2   

I downloaded the Homie zip from https://github.com/marvinroger/homie-esp8266/releases   which points to file https://github.com/marvinroger/homie-esp8266/archive/v2.0.0-beta.1.zip

Then i had the problem with latest ArduinoJSON incopatibility, the i downloaded the latest github version which corrected those errrors.

Now, i tried some older versions of ArduinoJSON but no luck, i still getting this error.
C:\Users\.................\Documents\Arduino\libraries\homie-esp8266-develop\src\Homie\Boot\BootConfig.cpp: In member function 'virtual void HomieInternals::BootConfig::setup()':

C:\Users\..............\Documents\Arduino\libraries\homie-esp8266-develop\src\Homie\Boot\BootConfig.cpp:48:39: error: passing 'const IPAddress' as 'this' argument of 'String IPAddress::toString()' discards qualifiers [-fpermissive]

   _apIpStr = ACCESS_POINT_IP.toString();

exit status 1
Error compiling for board WeMos D1 R2 & mini.

timpur:
[...]

PR #432, opened 2017-11-20
More Refactoring + Deep Sleep + Prevent WiFi Reconnect when reboot

timpur:
Main changes are:
- Small fix to String issue on config mode for older SDK versions (#429)
- Major Refactoring 
- Deep Sleep Function
- Prevent WiFi Reconnect when reboot


timpur:
not sure if this is correct... its showing its going to pull previous commits that have already been merged?

src/Homie.cpp
```diff
@@ -19,6 +19,7 @@ HomieClass::HomieClass()
   Interface::get().reset.triggerTime = DEFAULT_RESET_TIME;
   Interface::get().reset.resetFlag = false;
   Interface::get().flaggedForSleep = false;
+  Interface::get().flaggedForSleep = false;
```
luebbe: Any particular reason to do this twice?

timpur: It was an copy error. Fixed in the typo commit. Soz

luebbe: Just kidding, thx :)

timpur:
@marvinroger Ready for review. Want to push this through before working on more enhancements. 
Going to also rebase my fork after this pull request.

Whats Next:
- Add incremental config updates via config mode (Web UI and API) (#433).
- Add ability to remove settings via incremental updates.
- Fix incremental update trying to add settings when previously no settings object exists.
- Better event buss handler for internal events and external. Hopefully this will make long async possesses that occur in homie more readable, like the prep for deep sleep procedure.
- Better Task Manger for single shot async tasks + ongoing loop async tasks that occur throughout homie. Might also expose this for external use for the user to register tasks and loops. It will be something like the [TickerScheduler](https://github.com/Toshik/TickerScheduler) but custom to fit homie's needs.
- More refactoring.
- Possibly Upload Firmware via config mode (Web UI and API).

marvinroger:
@timpur thanks a lot! Can you please do a tiny commit to trigger a CircleCI build? I just reenabled it for PR. 😃

timpur:
@marvinroger dont know what to do about the other checks ? Think these need fixing. Also not sure about lint....

marvinroger:
I fixed the generate_docs step (I guess, it can only be tested on forks).

The linter is not happy because you do blocks that way:

```c++
if (1 + 1 == 2) {
}
else {
}
```

Whereas the Google style is:


```c++
if (1 + 1 == 2) {
} else {
}
```

Also, even though it might seem weird, we should not indent what's immediately inside a namespace "block". But don't worry, I'll take care of formatting the code. 👍 

Thanks!

timpur:
@marvinroger no i also want to get the code style correct so we need to fix this in the `.editorconfig` so auto format follows these styles. I'll look into the google style c++.

Looks like `.editorconfig` might not do what we want. Thus ill look into other solutions.

marvinroger:
@timpur editorconfig indeed only defines line endings, tabs/spaces, not much more. I am not sure there's a tool for that.

Just check out the changes I made: https://github.com/marvinroger/homie-esp8266/compare/e41bb9cd0aff069abe478379c4a109333dfdaaad...93a00e917b25c2a57d96b609fc0b2923981da7cd

timpur:
@marvinroger looks good. What are you then using for auto formatting?

marvinroger:
I don’t use any tools, I code that way. But that’d be an interesting
addition.
Clang might be suitable.

timpur:
@marvinoger do you want to release beta 3 since this fixes an issue with beta 2.

#429

bertmelis:
FYI, in VSCode and Atom you've got a cpplint extension.

tripflex:
@bertmelis yeah that's what I use, VSCode with PlatformIO 😛

PR #438, opened 2017-11-28
Simplify CI with @PlatformIO

ivankravets:


.circleci/config.yml
```diff
@@ -8,23 +8,20 @@ jobs:
       - checkout
       - run:
           name: install PlatformIO
-          command: sudo pip install platformio
-      - run:
-          name: install current code as a PlatformIO library with all dependencies
-          command: platformio lib -g install file://.
+          command: sudo pip install -U https://github.com/platformio/platformio-core/archive/develop.zip
```
marvinroger: Isn't it a bit dangerous to depend on develop? Shouldn't we pin to, I don't know, 3.5.0rc2?

ivankravets: It's OK, you will help us make PIO better if something will be wrong :) I mean, that is good that you will test your project with our development version. It's stable in development for 99%.

marvinroger: Alright, then. Thanks! :)

PR #439, opened 2017-11-28
Install library via @PlatformIO with all dependencies

ivankravets:
Sorry, now I understand why you re-install library from existing sources... Just to install dependencies. This is a good idea! Sorry again!


marvinroger:
Ahah, no problem. :wink:
Thanks for your help!

Issue #433, opened 2017-11-20
Feature Request +Config Mode can also edit current config


timpur:
Wondering if you guys think it would be nice to also use config mode to edit current configuration. Basically allow incremental json updates from api or web ui. This would be done by using the incremental json to update settings which already exists (from mqtt), so thats easy. 

Also thinking of changing the button to have two states to trigger a reset of config (current) or a restart into config mode to update current config. 
I was thinking of using https://github.com/mathertel/OneButton so we can use double click to put into update mode and long press like current to do full reset of config. This would basically be replacing https://github.com/thomasfredericks/Bounce2

Thoughts?

(Willing to implement by my self, just want to get ideas from you guys. Any additions?)

timpur:
@marvinroger @bertmelis @euphi @LiyouZhou @scharph @jesserockz

Gulaschcowboy:
Not being asked, but as a heavy homie-user: I vote +1

bertmelis:
I'm not waiting for it, but I not against neither.

But would it be still possible to run a webserver (on port 80) next to Homie?

timpur:
@Gulaschcowboy soz didn't mean to make it feel like i only wanted people who i tagged, not the case, just tagged them because they're contributes that i saw in recent commits. Anyone is always welcome to voice!!

@bertmelis not sure what you mean by this. Id be adding this feature to the homie web server and not subtracting anything from it? 

> But would it be still possible to run a webserver (on port 80) next to Homie?

bertmelis:
Maybe I don't get it completely.

Will the config-ui be available in config mode only or also in normal mode? As the latter will block any custom webserver that I may have implemented.

Or is the change only affecting config mode? (and the way you're getting into config mode)

marvinroger:
@timpur definitely, that's a great idea. Thank you so much for supporting this project, it's sad that I don't have enough time to work on it, but thanks to everyone involved in keeping this project alive. 😇

euphi:
OTA support would be nice, too. So you can update a firmware that crashes in NormalMode.

timpur:
@bertmelis change only to config mode to enhance the config web ui.

@euphi Use the config web ui to upload a new firmware? is what what you were thinking ?

luebbe:
How about removing config entries? Would this work (via incremental json) as well?

timpur:
@luebbe are you referring to #420, yeah id like to fix that issue for both mqtt and config mode.

[...]

Issue #420, opened 2017-11-08
How to Remove Homie Settings via MQTT


timpur:
Wondering how to use incremental json to remove settings from the device config?

stritti:
Seems to be stale issue.
Will be closed.
Feel free to reopen if there are more information.

Issue #372, opened 2017-07-23
Use Homie Settings before Homie Setup?


timpur:
Is it possible to use homie settings before homie is setup.

Id like it use this to set a parameter used fro advertising a node attribute (which has to be done before homie setup...).

Thanks Tim.

timpur:
Also wondering if its possible to set settings through code during running of homie with out going through mqtt?

euphi:
Yes, it is possible. But be careful when you use settings within a constructor (or anything called from a constructor) of a global object. See PR https://github.com/marvinroger/homie-esp8266/pull/335 for a solution to use HomieSetting in a constructor.

timpur:
@marvinroger, ideas?

marvinroger:
The problem is we're populating the settings objects at `setup` time. Introducing another method might be a solution, but we would have to refactor a bunch of things, but that's definitely a valid use case. Let's put this on 2.1.0 roadmap (as 2.0.0 is not even published final.. school, again and again 😕).

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

euphi:
Ah, yes. My PR #335 just makes it possible to add items to the vector  `IHomieSetting::settings` before `main()` (and thus the 'global' setup()) , but the value of the HomieSetting is read during `Homie::setup()`.

timpur:
> school, again and again 

I get that, in the same boat.

TBH dont know much about the internals of homie, ill slowly learn how it all works so i can make more meaningful contributions. Thus dont really know what you guys are talking about sorry :P

timpur:
Could we have a `Homie.loadSettings()` which loads the config? then use` Homie.setup()` like normal, but the load config part is optional thus, not interfering with the current workflow? (`Homie.setup()` would call `Homie.loadSettings()` if not been called, obviously...). 

Can you extract that part of the settings config code and use the settings objects like that, all before `Homie.setup()`?

euphi:
BTW: @marvinroger Can you merge my PR soon? It just changes a undefined (or random) behaviour to a defined one. 
I need the PR for my HomieNodeCollection, so for now I use my own fork, but this means to not update that often. It also complicates using platformio / travis.

luebbe:
I've beed struggling with this a lot. I'm subclassing Homie nodes which have custom properties. Example a Weather Underground Node that needs an API key, location etc.
[...]

PR #445, opened 2017-12-12
Feature Requests Towards v2.1

timpur:
## Feature Requests:
- [ ] Add ability to remove settings via incremental updates ( #420 ).
- [ ] Fix incremental update trying to add setting items when previously no settings object exists in the config.
- [ ] Add incremental config updates via config mode (Web UI and API) ( #433 )
- [ ] Better event buss handler for internal events and external. Hopefully this will make long async possesses that occur in homie more readable, like the prepare for deep sleep procedure.
- [ ] Better Task Manger for single shot async tasks (timeout) + ongoing loop async tasks (interval) that occur throughout homie. Might also expose this for external use for the user to register timeout tasks and loops. It will be something like the [TickerScheduler](https://github.com/Toshik/TickerScheduler) but custom to fit homie's needs.
- [ ] Add the ability to access settings before setup ( #372 )
- [ ] Add the ability to set the values of settings in homie and save to the config file ( #457 )
- [ ] Add possibly Upload Firmware via config mode (Web UI and API).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ The Git repository contains the development version of Homie for ESP8266. Stable
 
 1. Install [PlatformIO IDE](http://platformio.org/platformio-ide)
 2. Create new project using "PlatformIO Home > New Project"
-3. Open [Project Configuration File `platformio.ini`](http://docs.platformio.org/page/projectconf.html) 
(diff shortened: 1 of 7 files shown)


timpur:
Note: F() macro every string possible to save memory

PR #448, opened 2017-12-15
Depend on ArduinoJson@<5.11.0 via @PlatformIO

ivankravets:
Related issue https://github.com/platformio/platformio-atom-ide/issues/1249

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,9 @@ The Git repository contains the development version of Homie for ESP8266. Stable
 platform = espressif8266
 board = ...
 framework = arduino
-lib_deps = Homie
+lib_deps =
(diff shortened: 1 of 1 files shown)


timpur:
should work for 5.12.0 also.

euphi:
5.12 works  too. However, version 6 is out, which breaks compatiblitiy. But this needs to be fixed in library.json

JMacIV:
Hey @euphi - Using dev v3 branch and this issue still exists in library.json. Modified it to ArduinoJSON, version "">=5.10.0, <6"" and it worked again. Loses the compatible "^" operator though. Should I do a new PR for that branch?

Issue #457, opened 2017-12-22
Saving custom setting from device


tripflex:
As mentioned in my other issues/comments, I have a custom node that relies on a custom config value that is stored by Homie.

The issue is that, I also have exposed this setting in a node property, to allow configuring that one property via MQTT, but can't find any way to save a custom setting value (without using incremental json to do so).

Is there no way to set a custom setting value and have that save to the config file ... from inside my code base?  IF not, i'll just have to use a lib I created from my old project to save json values to a file ... would just be nicer to be able to save that value in the same homie config to not have to use multiple configuration files.

Case and point, I have a RelayNode (which controls a pump), and I have a custom setting setup for  `runDuration` ... this `runDuration` is also exposed as a property of the node, ie:
`/homie/test-device/pump/duration` which is settable, and when that value is set, I need it to update the `runDuration` saved to the HomieConfiguration, that way that value that was set is used instead of the default value.

Yes when the device connects back via MQTT and it subscribes it will receive the last set value, but if the device does not have internet, I still need it to be able to run and use the last set value, thus, the reason I need to be able to save the value to the configuration.

Is this something that would be possible already with the existing Homie codebase, or do I need to just go ahead and setup another way to handle saving my own custom configurations?

timpur:
Im working on alot to enable using settings better with homie, ill add this to the list ( #458 )

tripflex:
@timpur awesome 👍  thank you!!

timpur:
Added this to my fork. You can test it if you like

tripflex:
Yup I noticed that, i've actually been using your 2.1 branch fork for a while now, haven't had a chance to test this part yet, but will do so here this weekend and let you know

timpur:
@tripflex
Awsome, on hols for a week so will be smashing out some homie stuff while I have a bit more time... So report issues to me if they arise about V2.1

tripflex:
@timpur will do .. is the plan to keep the `/wifi/connect` as `POST`?  Reason I ask I create a Meteor plugin to handle Homie API (and anything else I can add later) and noticed you changed it to POST in your fork:
https://github.com/tripflex/meteor-homie

timpur:
Yes, I believe put should be for putting a file on the server and not setting a setting. So I think post makes more sense here

Issue #446, opened 2017-12-13
Configuration v2 fails due to CORS


jalmeroth:
I am trying to configure the latest dev-version of Homie 2 (f605adf).

When using the latest Configurator page v2 (03034c6) I am receiving an error in Chrome 63.0.3239.84 (Official Build) (64-bit) and other browsers:
`Failed to load http://192.168.123.1/heart: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://localhost:8000' is therefore not allowed access.`

I could temporarily fix this issue by adding the following default headers after this line https://github.com/marvinroger/homie-esp8266/blob/develop/src/Homie/Boot/BootConfig.cpp#L18:
```
DefaultHeaders::Instance().addHeader(F("Access-Control-Allow-Origin"), F("*"));
DefaultHeaders::Instance().addHeader(F("Access-Control-Allow-Methods"), F("PUT, GET"));
DefaultHeaders::Instance().addHeader(F("Access-Control-Allow-Headers"), F("Content-Type, Origin, Referer, User-Agent"));
```
_Please note: This is no the fix, it's just a proof to make it work temporarily._

Some of the following requests are also failing then as calls to __sendCORS() will send these headers again.

timpur:
Ill look into a fix for CORS, but adding those lines wont help you here, because your origin is from 8000 (Self hosted webserver?) not the esp. Thus you need 8000 to allow CORS not the esp, because you cant even get to the esp in the first place? (in my understanding,  might be wrong?)

Ill look into a solution for this. It would be nice to use the online config and not save to rely on the build in esp config, even though having it on the esp has other advantages, but thats up to you to decide and thats why ill try fix this.

jalmeroth:
> Ill look into a fix for CORS, but adding those lines wont help you here, because your origin is from 8000 (Self hosted webserver?) not the esp. Thus you need 8000 to allow CORS not the esp, because you cant even get to the esp in the first place? (in my understanding, might be wrong?)

That's wrong, `Access-Control-Allow-Origin: *` allows access from everywhere, including my local webserver (not the esp itself).

> Ill look into a solution for this. It would be nice to use the online config and not save to rely on the build in esp config, even though having it on the esp has other advantages, but thats up to you to decide and thats why ill try fix this.

Yeah, the online configurator is nice, but please keep the compatibility to checkout the gh-pages branch on your local maschine to configure homie devices, when you are offline or connected to the Homie Device AP. ;)

[...]

Issue #399, opened 2017-09-18
Renewing SSL certificate


sirvaulterscoff:
Hello,

i've got a question, which is not covered by current documentation. Let's say im using self-signed certificate with short-term validity - 2 month - for my homie-mqtt communication. Once certificate expires and should be renewed how am i supposed to update all my homie devices with new fingerprint ?

timpur:
Good point, i like, might add this to homie config file, thus you can just send incremental json to update it.
Will also implement homie ssl mode so its also easier to enable ssl.

timpur:
Has been added

stritti:
It is merged into `develop-v3`

Issue #320, opened 2017-03-18
MQTT not reconnecting after days of uptime vs. minutes / hours


johnjore:
Hi!

My mosquitto server goes offline from time to time (separate issue to resolve), but when this happens, homie does not always reconnect after the server is rebooted and mosquitto is again running. I can only reproduce the issue after letting it run for several days, not minutes or hours.

Here I left it running for a few days, and this morning I have this message (after the server crash):
> MQTT packet acknowledged, packetId: 7459
> ...
> Looping through the various sensor, but there has was no change in their state, until:
> ...
> Heap size: 30968
> Temperature: 24.20 °C
> Humidity: 43.20 %
> Lux: 21 lx
> Motion: CLOSED
> MQ:23.74    LPG:0.00 ppm    CO:0.00 ppm    SMOKE:0.00 ppm
> ✖ MQTT disconnected
> Triggering MQTT_DISCONNECTED event...
> MQTT disconnected, reason: 0
> ↕ Attempting to connect to MQTT...

After 8h in this state, its not moved on and it never tries to reconnect, but other sensors have. I've not managed to figure out what, or why, it does not reconnect. As I can see other sensors are reporting in, the problem is not with the mosquitto server. Log file on the mosquitto server shows no traffic from this node when it is in this state.

After resetting the device, and then manually stopping the mosquitto service:
> MQ:23.74    LPG:0.00 ppm    CO:0.00 ppm    SMOKE:0.00 ppm
> ✖ MQTT disconnected
> Triggering MQTT_DISCONNECTED event...
> MQTT disconnected, reason: 0
> ↕ Attempting to connect to MQTT...
> Retrying (500ms)...
> ↕ Attempting to connect to MQTT...
> Retrying (1000ms)...
> ↕ Attempting to connect to MQTT...
> Retrying (2000ms)...
> ↕ Attempting to connect to MQTT...
> Retrying (4000ms)...
> ↕ Attempting to connect to MQTT...

It all looks normal and when the mosquitto service is started, homie reconnects and all is good.

Versions: Using your mqtt client and homie (2) as of March 10th.

Any idea as to why it fails to reconnect after days of uptime vs minutes or hours?
Is there additional debug information I can provide?

JJ

marvinroger:
This one is going to be hard to debug. I'll try to run an ESP8266 for some days, but I am afraid the issue comes from the ESPAsyncTCP library (or maybe there's a memory leak somewhere).

johnjore:
[...]

PR #458, opened 2017-12-22
Towards V2.1

timpur:
- [x] Add ability to remove settings via incremental updates ( #420 )
- [x] Fix incremental update trying to add setting items when previously no settings object exists in the config.
- [x] Add the ability to access settings before setup ( #372 )
- [x] Add the ability to set the values of settings in homie and save to the config file ( #457 )
- [x] Get Device ID (#463)
- [x] TLS support + fingerprint attribute in config file ( #399 )
- [ ] Update Config UI for new features in the config file ( Also make it Smaller #469 ) (https://github.com/marvinroger/homie-esp8266-setup/issues/3) ( #446 )
- [x] MQTT Restart Device
- [ ] Fix MQTT retained messages ( #313 )
- [ ] Fix Bugs (#478 #320)
- [ ] Add incremental config updates via config mode (Web UI and API) ( #433 )
- [ ] Add possibly Upload Firmware via config mode (Web UI and API).
- [ ] Size of firmware ( #471 )
- [ ] Refactor Arduino Json to use latest lib to make config code more efficient. (https://arduinojson.org/api/jsonvariant/or/)
- [ ] Documentation 
- [ ] Update Examples
- [ ] Improve interface with '/set' prop


timpur:
@marvinroger @ThomDietrich @tripflex @euphi @bertmelis 

Will be changing from debounc2 to [One Button](https://github.com/mathertel/OneButton) this this will be more beneficial to homie and homie's user base. This will allow us to add a feature of changing boot modes via double click to swap over to config mode with out reset to update config via the ui or api.

It will also mean that users using buttons in there project will not need to look else where to get click, double click and long press functionality. This will also open up a handler for users to hook on to for normal button clicks (the rest used for homie) on the Homie reset button (GPIO 0)

Would like to know if there any major issues with this change, but its something i will go forward with for v2.1. I think this lib will serve us better. I know it will be breaking, but it is v2.1 so .....

kylegordon:
With regards to the One Button change, have you seen https://github.com/pkourany/clickButton at all? I currently use it with good effect for up to 3 clicks and 3 clicks and hold. From what I've seen, One Button only supports single and double click.

I can't vouch for the code at all, other than "it works" - https://github.com/kylegordon/homie-sonoff-touch

ThomDietrich:
Both are healthy since ~2014. The One Button seems to be more famous (Stars count) but clickButton offers more functionality. Did you compare code overhead?

timpur:
Defs will look into, thanks guys.

timpur:
@tripflex, about to add some new changes + some changes to the ui config site. This will allow homie to use the ui to update the the config and no need for resetting the device. Also some the ui can now support all config settings like manual ip, ssl ect. When i do release these changes can you also test them with me ?

kylegordon:
@timpur sorry, I missed your message a month ago :-(

The way I interpret it is that a 'double long press' (-2) is tap once, then tap and hold. triple being two taps followed by a hold.

I'll double check shortly and let you know if it's different

timpur:
Yeah for me when testing that's how I got to -2 and -3, thought that was a bit odd and not what I was looking for ....

timpur:
@tripflex use the new ui to test out patch updates to the config 
https://github.com/timpur/homie-esp8266-setup/blob/revamp/ui_bundle.gz

luebbe:
Concerning the short and long presses: I implemented something similar a while ago. The short presses select the function and the long press confirms it. If you don't confirm within a second, the state switches back to "nothing".
Example: 
* 2x short + 1x long -> activate function two
* 3x short + 1x long -> activate function three
* 3x short -> nothing happens
* 1x long -> nothing happens (unless you implement a function zero)

So there's no Morse "short short long long short" possible :)

timpur:
My issue is how do I get 3x long 
It doesn't seem to work with that lib clickButton

luebbe:
I understand your problem. But is it really a problem in terms of usability? Isn't it easier to remember: "Reset config -> function 3 -> 3x short 1x long to confirm" instead of "Reset config = long long long", "Function xyz = long short long"?

euphi:
@timpur Do you know, why this PR was never merged? I try to get an overview of the changes, but these are too many to understand them soon :scream: . However, it seems at least some of them are useful also for #develop-v3 branch.

timpur:
@euphi think i was to ambitious with the task list :P

Maybe best to port some of the changes to the new branch ... ?

euphi:
Interestingly, therre is no merge conflict, although there were some many other commits on the #develop branch.

euphi:
@timpur Can you help with selecting the important commits? (Or help with testing. especially the changes to config mode, because I don't use the config mode).

timpur:
I will try help out when I can :)

Though I welcome others to also Chery pick commits into smaller PRs. I should of done this ...

stritti:
I close this PR. Most features are implemented on #develop-v3 now.

PR #460, opened 2017-12-26
Add last step to uibundle README

ThomDietrich:


timpur:
@ThomDietrich remember, people are free to use what ever IDE they want which means we need to make it as generic as possible, or support the main IDE's.

PR #462, opened 2017-12-26
Unable to use both SPIFFS and SD

awigen:
The Arduino core does not allow both SPIFFS and SD to be used at the
same time.

Ref: https://github.com/esp8266/Arduino/issues/3710

### DIFF
--- a/src/Homie/Config.cpp
+++ b/src/Homie/Config.cpp
@@ -27,7 +27,7 @@ bool Config::load() {
     return false;
   }
 
-  File configFile = SPIFFS.open(CONFIG_FILE_PATH, "r");
+  fs::File configFile = SPIFFS.open(CONFIG_FILE_PATH, "r");
   if (!configFile) {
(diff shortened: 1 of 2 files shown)


timpur:
Ill need to test this and check this is the best method, but looks great, thanks. 
Ill get on to this soon.

awigen:
FYI, the reason this is required is to support displays from 4dsystems.com.au like this one:
http://www.4dsystems.com.au/product/gen4_IoD/

These displays store graphics on a SD card while also requiring access to the SPIFFS flash. So it would be great to get this added before v2.0 final.

timpur:
it will be added (next free time i have soz)

timpur:
I have added this to v2.1 on my fork, Thus this will be included with v2.1 release.

(there have been quite alot of config changed and tbh cbf to merge this via git ....... so i just did it in v2.1, hope your okay with this).

PR #468, opened 2018-01-03
Correct API docs for /wifi/connect from GET to PUT

tripflex:
Docs show `/wifi/connect` as `GET` when it should actually be `PUT`

### DIFF
--- a/docs/configuration/http-json-api.md
+++ b/docs/configuration/http-json-api.md
@@ -158,7 +158,7 @@ If anything goes wrong, a return code != 2xx will be returned, with a JSON `erro
 
 --------------
 
-??? summary "GET `/wifi/connect`"
+??? summary "PUT `/wifi/connect`"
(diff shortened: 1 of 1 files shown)


timpur:
hmm thinking about this now `wifi/connect` should be a post not a put, since were updating the connected wifi and not creating a new resource.

tripflex:
Yup no problem, just wanted to make sure docs were correct.  I've forked this npm module https://github.com/tripflex/node-homie-config and will be updating it accordingly, as well as creating a Meteor package for it as well, so **PLEASE** tag me in any commits when you update this, so I can make sure that the config npm module and meteor package are correct

timpur:
@tripflex you know you can also upload the config.json directly onto the module from the computer ? just like flashing. This is be quickest method ..... and most repeatable .....

> It seems the only options are a web-based interface or an Android app. I wasn't very happy with this so I created a node.js module so I could configure one or more boards easily.

Love the node module, just making sure you know all the options ???

PR #484, opened 2018-02-02
Hint the use of a macro for fw version number

ThomDietrich:
Not being able to simple insert a variable as parameter to the function `Homie_setFirmware` comes as a surprise. After checking the macro behind it, I believe this is the cleanest option for outsourcing of the firmware version - which you might want to do in bigger programs.

I wasn't sure where to put this info, decided to go with one of the basic examples and the api reference documentation. Better ideas?

### DIFF
--- a/docs/others/cpp-api-reference.md
+++ b/docs/others/cpp-api-reference.md
@@ -27,7 +27,10 @@ void Homie_setFirmware(const char* name, const char* version);
 // This is not a typo
 ```
 
-Set the name and version of the firmware. This is useful for OTA, as Homie will check against the server if there is a newer version.
+Set the name and version of the firmware.
+This is useful for OTA, as Homie will check against the server if there is a newer version.
+Be aware, that the function is implemented as a `define` macro.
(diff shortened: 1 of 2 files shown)


luebbe:
I'm afraid that this project is in dire need of an active maintainer...

euphi:
> I'm afraid that this project is in dire need of an active maintainer...

@marvinroger , @timpur  What do you think about moving the repository to the homieiot organization?
Also, we should create a branch for homie-eps8266 v3.0 development (core feature: support of convention 3.0 and support for ESP32) .

BTW: I was inactive for the last months because I was very busy with other projects, especially the relocation of our local fablab (where I'm in the council and the need for relocation was quite short-termed and unexpected). Good thing is that the relocation is done and I now take care of the automation of the lab and I try to convince the others of using Homie where possible (vs. Tasmota as "competition").

And best way to convince is to provide software that is running stable ;-)

bertmelis:
maybe also move the asyncmqttclient to the organisation.

stritti:
I think, there is no question for merging this ...

Issue #487, opened 2018-02-11
loopHandler not called until Homie is connected


jhergeth:
My device registers a loopHandler with setLoopFunction() and does all its work within that loopHandler.
Now I want my device in a "boot" mode until the WLAN is connected.
Having this code:
```
void loop() {
    if(Homie.isConfigured() && !Homie.isConnected()){
        layt->setBootMode(bm);
        loopHandler();  // this is neccessary !!!
    }
    else{
        layt->setBootMode(0);
    }
    Homie.loop();   
    delay(1);
}
```
I needed to add the line with "loopHandler()" in the main loop to be called when the WLAN is not connected in order to have my code in loophandler() beeing called.

In my opinion Homie.loop() should call the handler which has been set using setLoopFunction() in all cases. Or, did I miss something in the documentation?

With regards
Joachim

cecilios:
This is related to #354

Nzbuu:
I just put the code that I want to run always in the loop function and the code that requires an MQTT connection in the loop handler.

euphi:
Closed with #490.

Issue #354, opened 2017-06-08
Homie 2.0: Create an option for boot normal mode


cecilios:
I am building an IoT irrigation controller for the garden, based on MQTT. It is, basically, an standalone device that should work by itself when the network is not available. Of course the design expects a network and all interaction with the irrigation controller (i.e. programming irrigations, setting times, etc.) is done via MQTT and needs network connection. But it the network fails, the controller must continue doing its job and all scheduled irrigations should be done.

I have in operation in my garden an initial version of this irrigation controller and I was trying to implement OTA update when I found your Hoime project. By the way, Homie is an excellent project!! Congratulations!. As Homie fits perfectly for all other devices I have to develop my plan is to use Homie for them. And so, I decided to study the convenience of preparing a version of my irrigation controller based on Homie. It will cost me more or less the same than developing my own OTA update code. Therefore as my controller is a class, I derived it from HomieNode and removed some unnecesary services (basically dealing with network and MQTT connections). In a couple of days the irrigation controller was operational as a Homie node.

Now I am evaluating if definitelly the irrigation controller device should follow this approach (the irrigation controller as a Homie node) or better continue with my previous approach (code independent from Homie, but MQTT messages compatible). The main drawback for moving the irrigation controller to Homie is that in Homie the nodes in a device do not receive service (their loop method is not invoked) when the network connection is lost. And this is a serious problem for me, as LAN connectivity can be lost by many reasons but the important thing is that irrigations must continue according to the scheduled programs. Network connectivity is just a commodity for supervising the device operation (feedback received via MQTT) or for changing operational parameters via MQTT (setting irrigation programs, times, etc.). Not having connectivity it is not a serious problem for stopping irrigations.

The Standalone boot mode is not of use in this case, as it assumes no network connection. 

What I really need is that in normal boot mode, in case of no connectivity, the Nodes::loop() method should be invoked so that the irrigation controller can do its job (but as there is no connectivity it will not send MQTT feedback to the domotic controller).

This would require a modification in BootNormal.cpp, just to remove the loop:

   ```
 for (HomieNode* iNode : HomieNode::nodes) {
      iNode->loop();
    }
```

from inside the if statement:
 
 ```
 if (Interface::get().connected) {
  }

```

But as this would create incompatibilities with existing software using Homie, I would suggest to create either a new boot node (NORMAL_STANDALONE) or an option for boot normal method. Would that be posible?

Thank you.

jamesgol:
[...]

PR #490, opened 2018-02-13
Proposal to optionally run HomieNode::loop() also in disconnected state

euphi:
See #354, #487


src/Homie/Boot/BootNormal.cpp
```diff
@@ -97,6 +97,9 @@ void BootNormal::loop() {
     ESP.restart();
   }
 
+  for (HomieNode* iNode : HomieNode::nodes) {
```
Nzbuu: Why is this moved? It may break functionality that assumes that loopFunction is executed before the individual node loop methods?

Nzbuu:
I'm strongly opposed to this implementation. It could break existing HomieNodes that assumes that the loop methods are only executed when MQTT is connected. Furthermore, if a particular HomieNode expects the runLoopDisconnected to be set to function correctly, then there is no way to enforce that and users may forget to set the flag, creating unexpected behaviour.

I suggest to have two "loop" methods per HomieNode: one that only runs when connected (as currently) and one that runs even when disconnected. Then it is up to the implementation of each HomieNode to determine when the loop functionality should run.

src/Homie/Boot/BootNormal.cpp
euphi: After the loop is before the loop.

Prior the change, the loop-function was called as last. Now it is called first, so the order has not changed.

euphi:
runLoopDisconnected  is initialized to false, so the default behaviour has not changed.

Furthermore, runLoopDisconnected  is a field of HomieNode, so each HomieNode can set the flag itself. It even couldoverride the setter, so no one else can change this.
 (I think it is sufficient that the setter is non-virtual, because no user would access a derived class as HomieNode unless he knows what he's doing - however, if you think that this is not enough, the setter can be made virtual).

src/Homie/Boot/BootNormal.cpp
Nzbuu: OK. Maybe it's not so bad, but the end of the loop and the start are only the same if there's nothing else in the Arduino loop function.

Nzbuu:
OK. I take back my comments about runLoopDisconnected. I was a bit hasty. Although, I think that the setter should be protected rather than public, since the Node constructor should set it and never touch it again.

I still think it would be better with two loop methods for the Node, one which runs always and one which runs when connected.

timpur:
Compermise ? Pass is connected into the loop ?

euphi:
@Nzbuu you're rigth with the setter. There is no need for it, if you can set the flag in constructor. I will add an optional parameter to `HomieNode::HomieNode(..)`.

euphi:
Oh, and I don't see the benefit of passing a connected parameter to loop(), because it is easy to "ask" homie for it's current state.

timpur:
More for convenience

But if this loop is running at max speed then asking for the status every look could slow down the loop speed ???

stritti:
Would be cool to have loops which are running without connection to build more robust controllers which are working without connections too.

Any progress on this issue?

euphi:
As this is for the development-version (both v2 and v3), I think the minor risk of side effecs due to the different running order in the loop is acceptable.
This closes #571, #536, #487, #354

kleini:
For this breaks all my existing HomieNodes. Previously I could assume that MQTT is ready when the HomieNode::loop() method is called. This is not the case any more and results in MQTT problems. My HomieNodes read sensor values in the loop method() and send them via MQTT. MQTT sending does not work any more as MQTT ready is not reached and SendPromise::send() refuses to send data if MQTT ready is not reached.
For me this merge has a big impact and breaks all my HomieNodes although runLoopDisconnected is still false in my cases.

kleini:
So my question is, is this intended that HomieNode::loop() should be executed although MQTT ready is not reached and SendPromise::send() refuses to send messages? Do I need to check manually for MQTT ready to make send() not fail?

PR #496, opened 2018-02-23
add a feature called nestednode

erow:
Add ability to create a new device by combining  devices which already have .


erow:
Maybe apply this feature as a plugin is a better way.

timpur:
Lost on what this is ?

euphi:
Eh.. you only commited a lot of cosmetical and editoral changes and deleted almost all examples?

But the important and interesting part is missing (especially the files NestedNode.hpp / .cpp)?!?!

timpur:
Yes, this makes no sense, what is nested nodes ?

timpur:
This goes against the homie convention and will not be implemented. That said the idea of predefined nodes has been discussed as a feature that will come in the form of addons in a separate repo.

erow:
You have seen my codes? I made some mistakes that cause many problems. My IDE enable auto-format, and change lots of unnecessary files.

> Homie Convention 3.0 (redesign)

Where can I find it?

timpur:
https://github.com/marvinroger/homie/tree/redesign

erow:
That's more impressive.But I still don't like the idea of `Property attributes`.It limits what we send.'nested nodes'  allow developer write complex json message to control device.And can control multi devices simultaneously or part of device you need, that's my favorite function.

timpur:
Thats fine, but still dont think it should be part of the core, the core should follow the Homie Spec and thats about it. Need to keep homie from getting bloated, thus we will be developing additional things as addons. This would be nice as a addon.

erow:
I agree that. So I create a new project [NestedNode](https://github.com/erow/NestedNode.git). There are still some problems, but it shows what 'Nested' means.

timpur:
Love it, options are always welcome. Will soon create an add-ons doc page for homie where we can feature things like this.

PR #507, opened 2018-03-26
Update to Homie Convention v2.0.1

timpur:


src/Homie/Boot/BootConfig.cpp
```diff
@@ -335,7 +335,7 @@ void BootConfig::_onDeviceInfoRequest(AsyncWebServerRequest *request) {
   for (IHomieSetting* iSetting : IHomieSetting::settings) {
     JsonObject& jsonSetting = jsonBuffer.createObject();
 
-    if (String(iSetting->getType()) != "unknown") {
+    if (strcmp(iSetting->getType(), "unknown") != 0) {
```
euphi: Why strcmp?

timpur: In the core of the lib, we should try avoid using String. String will copy the chars into dynamic memory even for something as simple as this... Using String everywhere can lead to heep fragmentation.

bodiroga:
Hi @timpur!

Do you plan to include this PR in the final Homie-ESP8266 V2.0 version?

Thanks!

timpur:
Not sure, what do you guy think ?

euphi:
Yes, of course. It is ready now, so it should be added.

PR #512, opened 2018-04-02
Add option for ota_updater script to connect to TLS-protected brokers

insertjokehere:
My MQTT broker refuses connections that aren't made over TLS, so I can't use the ota_updater script as-written. This PR adds a `--broker-tls-cacert` that enables TLS for the connection and sets the the CA cert bundle used to validate the connection

Additionally, this fixes a bug where the updater wouldn't work if the target device was asleep


insertjokehere:
It works for me™

Issue #524, opened 2018-06-01
Subscribing to messages using Homie.getMqttClient() is not reliable: topic is truncated


peret2000:
Hello,

I am trying to use Homie as my base code for all the developments in my ESP8266. It is a great framework to run projects!
In my code, I am subscribing to messages from other Homie devices: let's say device 'A' subscribes to messages from device 'B' (homie/B/#).

It is done as this example:

```
_AsyncMqttClient& mqttClient = Homie.getMqttClient();                                                                                                                 
uint16_t packetIdSub = mqttClient.subscribe("homie/B/#", 2);
...
mqttClient.onMessage(onMqttMessage);
...
void onMqttMessage(char* topic, char* payload, AsyncMqttClientMessageProperties properties, size_t len, size_t index, size_t total)
{                                                                                                                                                                      
  Serial << "[MQTT onMqttMessage]: " << topic << ":" << ((String)payload).substring(0,len) << endl;
}

```
I have noticed that topic is truncated to 'homie/B'

The reason, as far as I have understood the code, is that BootNormal::_onMqttMessage(), that is the first event handler called by async-mqtt-client library when a subscribed message arrives, modifies 'topic'
This method calls to __splitTopic(), that calls to strtok(), to split the topic, and fill in _mqttTopicLevels.
That function exchanges '/' with a '\0' character after its first token (in my example 'homie/B').

Later, when you deal with 'topic', in onMqttMessage() , it has been truncated, because you are called after BootNormal::_onMqttMessage().

I see, at least, two solutions:
- One, more transparent, would be to make a copy of 'char* token' at the begining of BootNormal::_onMqttMessage(), and work with the  copy. I need more experience in programing to do that, as I am not aware of all of the implications of allocating memory in a library that I do not fully know.

- And two: BootNormal::_onMqttMessage() could call first to the rest of functions that are subscribed to mqtt messages, and then, continue to do its job. This will probably imply more changes in code

Homie library would be much more powerful if we can use devices for generic purposes, based on Homie, not restricted to some 'limitations'. But that may need changes, like the ones proposed here.

I am sorry if all of this is wrong, because I am not an expert, maybe I am not coding as I should. In such case, please ignore this comment.

Regards

timpur:
@peret2000, you make a good point. You should be able to tap into the mqtt lib and use it as expected it you really need to. I shall fix this/if you want to you can.

[...]

PR #526, opened 2018-06-03
Subscribing to messages using Homie.getMqttClient() is not reliable: topic is truncated

peret2000:
It is not possible to subscribe to MQTT messages in general using Homie framework, as BootNormal::_onMqttMessage() truncates the topic.
[Issue #524](https://github.com/marvinroger/homie-esp8266/issues/524) 

With this change, topics can afterwards be intercepted by any other subscribed function, without loosing functionality

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -20,8 +20,9 @@ BootNormal::BootNormal()
   , _mqttWillTopic(nullptr)
   , _mqttPayloadBuffer(nullptr)
   , _mqttTopicLevels(nullptr)
-  , _mqttTopicLevelsCount(0) {
-  strlcpy(_fwChecksum, ESP.getSketchMD5().c_str(), sizeof(_fwChecksum));
+  , _mqttTopicLevelsCount(0)
+  , _mqttTopicCopy(nullptr)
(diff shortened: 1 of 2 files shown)


peret2000:
I tested it, seemed correct. Unfortunately I do not have any test environment nor something with robust tests; I just tested with some code examples

sradner13:
Updated my environment (develop-v3) with this change and OTA updates started to encounter very inconsistent watchdog timer errors.  With this change backed out, I can once again reliably update.  I believe the extra copy causes enough of a delay to make the script unstable.  If others have encountered problems recently with OTA updates, I would be curious if this change is suspect.

Sorry I do not have an alternative fix to the original problem.

sradner13:
A few more details if this helps anyone.  I'm not sure the right issue to be updating, so apologize if this is in the wrong place.

My hardware is ESP8266 based MakerFocus D1, Mini, which seems to be very similar to the Wemos D1 Mini.
I'm building for: ESP-12E / 4M: 1M spiffs / LwIP 2.0 higher bandwidth.

Since backing out this change I have been very successful with OTAs (about 12 tests with no failures).  It also appears the boards are more reliably keeping up MQTT connections to the broker.

While I do not have another code solution for the original problem, I'm happy to help test if someone has a proposed fix.

Thanks for the awesome library.

Issue #532, opened 2018-07-03
OTA does not work if TLS/SSL is enable


jmozmoz:
If TLS/SSL is enabled by the following entries in main.cpp, config.json and platformio.ini, OTA is not working. If SSL is disable, OTA is working.

main.cpp
```
#if ASYNC_TCP_SSL_ENABLED
  Homie.getLogger() << endl << endl << F("ESP Homie Using TLS") << endl;
  Homie.getMqttClient().setSecure(true);
  Homie.getMqttClient().addServerFingerprint((const uint8_t[])MQTT_SERVER_FINGERPRINT);
#endif
```

config.json
```
  "mqtt": {
    "host": "xxx", 
    "port": 8883,
    "base_topic": "homie/",
    "auth": true,
    "username": "aaa",
    "password": "bbb",
    "ssl": true,
    "ssl_fingerprint":"abcdef12345"
  },
```

platformio.ini
```
[env:esp01_1m]
platform = espressif8266
board = esp01_1m
framework = arduino
;build_flags = -Wl,-Teagle.flash.1m64.ld
build_flags = -DASYNC_TCP_SSL_ENABLED=1 -Wl,-Teagle.flash.1m64.ld
lib_deps =
  https://github.com/marvinroger/homie-esp8266.git#v2.1
  simpleTimer
  jled

; Serial Monitor options
monitor_speed = 115200
```

Disabling SSL is done by changing the port to 1883 and removing the -DASYNC_TCP_SSL_ENABLED from build_flags

The output in the serial console, if SSL is enabled and an OTA is send to the device is just:
```
✖ MQTT disconnected
Triggering MQTT_DISCONNECTED event...
↕ Attempting to connect to MQTT...
Sending initial information...
✔ MQTT ready
Triggering MQTT_READY event...
〽 Sending statistics...
  • Wi-Fi signal quality: 96%
  • Uptime: 114s
```
and the ota_updater.py just gets back the original checksum
```
Connecting to mqtt broker alpha on port 1883
Connected with result code 0
Waiting for device info...
Publishing new firmware with checksum 6a0c1c4c31e15b2c7cd6970ffdea8cb3
Expecting checksum 6a0c1c4c31e15b2c7cd6970ffdea8cb3, got 5a073a977e0536f7651faa0c8d77ca39, update failed!
```

timpur:
Please see #458

I don't believe TLS has been pushed to here yet, it's only in my fork branch for v2.1

[...]

PR #547, opened 2018-10-24
Require use of PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY

sillyfrog:
A small documentation update so users set the `PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY` build flag for reliable OTA updates.


jmozmoz:
Have you checked, if SSL connections are still working after applying this patch? In my tests (see https://github.com/marvinroger/homie-esp8266/issues/532#issuecomment-425743469) `PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY` breaks it.

Issue #549, opened 2018-11-04
Homie 3.x support


davidgraeff:
The convention has evolved and I wonder if this original implementation that developed the convention will support it at some point.

euphi:
Merged. Please support further development by testing and improving :-D

Issue #467, opened 2018-01-02
Homie 3.0


timpur:
Open to any one.

Like to discuss what you think homie 3.0 will be, new feature, more tools, ext. What would you like to see.

Some of my thoughts.
- Under to hood (core changed).
    - New event bus.
    - New task manager.
    - Possibly a add-on aritecture. So you can enable and disable parts of homie core as you need them.
- Conform to Homie 3.0 (TBA)
- Add-ons
    - Battery management
- Compatibility with esp32???(not sure about this yet)

Let's talk.

furyfire:
Streamline the usage of JSON. Currently the $properties topic is comma seperated, but the $implementation/config topic is JSON. I don't care which one it is. I just want it to be streamlined. 

Task management is a big plus. Some sort of repository of nodes which you can easily pull from and quickly modify. (Read X sensor on pin Y every Z seconds)

Core addons sounds great if the implementation can be kept clean
Add on ideas
- Battery (readbattery level, sleep intervals etc)
- Debugging (Push debug output over internet, accept external commands etc)
- FileSystem (Push/Pull files to the SPIFFS over MQTT)
- Webserver in normal mode (For advanced user configuration)
- Other IoT protocols implementations (Philips Hue, Alexa, Google Iot, ThingSpeak)

I know these things could be implemented as nodes. I just feel like they would belong better as "add-on" modules that provide offical functionality in line with the Homie protocol.

IVData:
Is SSL support a possibility for this?

timpur:
@furyfire love it. Thank you for your comment.

euphi:
Regarding Add-Ons/Nodes: 

I think most addons will be somehow a HomieNode. So, most work for addons will be to establish a add-on database and an offiicial process how to add the addons there. 

For some cases a extension of the Homie API to support "Add-Ons" will be necessary, e.g. access to write "device properties" for the Battery add-on or more level of property information, e.g. for the FileSystem.


Regarding Debug: I have a [LoggerNode](https://github.com/euphi/HomieLoggerNode) that could be a good starting point for a debug add-on. Until now, it is not directly used by Homie, but it will be easy to adapt, especially if there would be a specific way to set a debug-functions inside homie.
(Or just subclass the Node also from `Serial`).

timpur:
[...]

Issue #464, opened 2017-12-30
Adherence to Homie Convention?  (lack of attributes)


awilliam:
I'm entirely new to this, so I apologize in advance if I'm misunderstanding.  Reading the Homie convention (https://github.com/marvinroger/homie) and comparing to the TemperatureSensor example, I see:
  temperatureNode.advertise("unit");
which is later set in the setupHandler as:
  temperatureNode.setProperty("unit").send("c");
Monitoring the MQTT traffic from the device, I see:

homie/000001/temperature/$properties', ... (12 bytes))
unit,degrees
homie/000001/temperature/unit', ... (1 bytes))
c
homie/000001/temperature/degrees', ... (5 bytes))
22.00

As I read the convention, 'degrees' should be a property on the node and '$unit' should be an attribute of that property.  These are not separate properties, one is an attribute of the other and $unit should appear as:

home/000001/temperature/degrees/$unit

(and should probably be '°C' rather than simply 'c', nit)

How do I use homie-esp8266 to properly create attributes on properties?  Additionally, how do I even set the $name attribute on a node as the interface only seems to allow setting the $type in the constructor and $properties via setProperty()?  Thanks!

bertmelis:
The current esp8266 implementation is not yet conform latest (and not yet final?) convention.

Work in progress.

awilliam:
Ok, well that's confusing and disappointing and really seems like it should be mentioned in the known limitations in the docs if not the API reference itself.

timpur:
Yes the convention has recently been changed and has not yet be released. When it is it will be implemented in homie 2.2.0

timpur:
Actually version names are still undecided. But it will be happing soon. I'm just waiting for final release of the new conversation then I'll implement.

stritti:
I think we could close. V3.0.0 is supported now.

lhanson:
@stritti Is this documented anywhere? As it stands I don't see any way to set property attributes. I can have a "temperature" node which advertises "units" as a property, but no way to set units as an attribute per the spec. Maybe I'm missing something?

MajorTwip:
https://homieiot.github.io/specification/spec-core-v3_0_0/#property-attributes

lhanson:
I'm aware that the spec includes property attributes, which is why I'm asking about them. I don't see any functionality in `homie-esp8266` for setting them. From the example in the spec:

```
homie/super-car/engine/temperature/$name → "Engine temperature"
[...]

PR #550, opened 2018-11-05
Adapt library to Homie 3.0.0

bodiroga:
Hi @marvinroger and @timpur!

Here you have my proposal to adapt the library to the newest Homie convention version :+1: I have tested the changes with a NodeMCU board and the provided examples compile successfully! Anyway, review the changes carefully, please, I'm not a C++ guy.

Some things that could be added in a future PR:

- Add a new method to name each node array element individually (right now, the "node name" is the same as the "node id" -> strip_1, strip_2, strip_3 and strip_4 in the "LedStrip.ino").
- Rework the "node" and "property" creation. For the "property" attributes, instead of providing all fields  in the "advertise" function as parameters, add individual functions and chain them. For example: myNode.advertise("temperature").withName("Temperature").withDatatype("int").withUnit("ºC").withFormat("-5:50").settable(myFunction). Even if it is more verbose, IMO it is more readable and users don't have to remember the order of the parameters. Also, it makes easier to add more fields, because we don't have to overload the function signature.

Many thanks for all your amazing work and don't hesitate to criticize my code, feel free :wink: 

Best regards,

Aitor

PS: I have left all the individual commits in case you want to review them one by one, but I can squash them into a single one for the final merge.
PS2: related issues: #549, #467, #464


bodiroga:
Hi @marvinroger and @timpur!

I'm going to add the "$retained" property attribute to conform to Homie 3.0.1 and I see this as the perfect opportunity to refactor the "advertise" function.

Do you agree that the best option is to define the advertise function with the only required attribute (id) and add the optional ones (name, unit, format, datatype, retained and settable) following a "builder" pattern? We actually do that with the settable attribute, so it's natural to extend that principle to the rest non-mandatory attributes. What's your favorite name? ".withName", ".name", ".setName",...?

Many thanks for you help and looking forward to hearing from you guys!

Aitor

src/HomieNode.cpp
```diff
@@ -9,8 +9,29 @@ PropertyInterface::PropertyInterface()
 : _property(nullptr) {
 }
 
-void PropertyInterface::settable(const PropertyInputHandler& inputHandler) {
+PropertyInterface& PropertyInterface::settable(const PropertyInputHandler& inputHandler) {
   _property->settable(inputHandler);
+  return *this;
+}
+
+PropertyInterface& PropertyInterface::setName(const char* name) {
```
marvinroger: I'm fine with the `setXXX` syntax, that way, it stays consistent. :)

bodiroga: Great Marvin!

What about the name of the function to set the retained attribute to 'false' (as it is 'true' by default)? ".retained(false)", ".non-retained()"? Maybe something non-mqtt related like ".event()"?

bodiroga:
Done, Homie 3.0.1 fully supported!

timpur:
sorry need to still have a proper review ....

src/HomieNode.cpp
davidgraeff: Within the convention the decision was to stick to the MQTT defined wording. For consistency, just call it "setRetained()".

examples/LedStrip/LedStrip.ino
```diff
@@ -1,9 +1,9 @@
 #include <Homie.h>
 
-const unsigned char NUMBER_OF_LED = 4;
+const unsigned int NUMBER_OF_LED = 4;
 const unsigned char LED_PINS[NUMBER_OF_LED] = { 16, 5, 4, 0 };
 
-HomieNode stripNode("strip", "strip");
+HomieNode stripNode("strip", "Strip", "strip", "true", 1, NUMBER_OF_LED);
```
davidgraeff: The constructor signature is `HomieNode(const char* id, const char* name, const char* type, bool range, uint16_t lower, uint16_t upper, const NodeInputHandler& inputHandler)`
Shouldn't that just be `true` instead of `"true"` for the type?

src/HomieNode.cpp
```diff
@@ -9,18 +9,48 @@ PropertyInterface::PropertyInterface()
...
+  return *this;
 }
 
 PropertyInterface& PropertyInterface::setProperty(Property* property) {
   _property = property;
   return *this;
 }
 
-HomieNode::HomieNode(const char* id, const char* type, const NodeInputHandler& inputHandler)
+HomieNode::HomieNode(const char* id, const char* name, const char* type, bool range, uint16_t lower, uint16_t upper, const NodeInputHandler& inputHandler)
```
davidgraeff: In all your examples the "id" is the same as "type".
"type" is actually optional, at least it doesn't say anything about the node in the controller perspective. 
Could you maybe remove it from the constructor and make it a chainable setter instead?

davidgraeff: Make all the array specific parameters a dedicated setter as well (`setArray(uint16_t lower, uint16_t upper)`). That way if arrays change at some point, we don't need to change the constructor.

src/SendingPromise.cpp
```diff
@@ -50,16 +50,16 @@ uint16_t SendingPromise::send(const String& value) {
...
   strcat(topic, _node->getId());
-  strcat_P(topic, PSTR("/"));
-  strcat(topic, _property->c_str());
-
   if (_range.isRange) {
     char rangeStr[5 + 1];  // max 65536
     itoa(_range.index, rangeStr, 10);
     strcat_P(topic, PSTR("_"));
     strcat(topic, rangeStr);
   }
```
davidgraeff: Ranges are not part of Homie 3?

examples/GlobalInputHandler/GlobalInputHandler.ino
```diff
@@ -1,8 +1,8 @@
 #include <Homie.h>
 
-HomieNode lightNode("light", "switch");
+HomieNode lightNode("light", "Light", "switch");
 
-bool globalInputHandler(const HomieNode& node, const String& property, const HomieRange& range, const String& value) {
+bool globalInputHandler(const HomieNode& node, const HomieRange& range, const String& property, const String& value) {
```
davidgraeff: Ranges are not part of Homie 3?

examples/LedStrip/LedStrip.ino
bodiroga: Absolutely correct :+1:

src/HomieNode.cpp
bodiroga: > "type" is actually optional, at least it doesn't say anything about the node in the controller perspective.

Type is a required attribute for Nodes in Homie 3.0.1, see here: https://homieiot.github.io/spec-core-v3_0_1/#node-attributes. That's why I have added it to the class constructor. Perhaps we can remove it for Homie 4.0, but that discussion doesn't belong here :wink: 

> Make all the array specific parameters a dedicated setter as well (setArray(uint16_t lower, uint16_t upper)). That way if arrays change at some point, we don't need to change the constructor.

I didn't want to change that part of the code too much, because it was already implemented, just the minimum necessary changes. What do you think @marvinroger?

src/SendingPromise.cpp
bodiroga: Ranges are how Node arrays are called in homie-esp8266, it's just an interval attribute name.

examples/GlobalInputHandler/GlobalInputHandler.ino
bodiroga: Ranges are how Node arrays are called in homie-esp8266, it's just an interval attribute name.

jmvaz:
Hi! When PR will be merged?

GideonLeGrange:
I wish we could get this merged and released 😕

 I'm running my fork of @bodiroga 's develop-v3 branch and it's fine. ESPs running this code integrates easily into Openhab 2.4.

jmvaz:
Come on guys! A little Christmas gift :)

manswiss:
Please merge it. I usw actualy this Form and work perfect.
For OpenHAB 2.4 is Homie 3.0 nessesaire.
Thanks.

src/HomieNode.cpp
```diff
@@ -37,28 +67,34 @@ HomieNode::~HomieNode() {
...
-  return _propertyInterface.setProperty(propertyObject);
+SendingPromise& HomieNode::setProperty(const String& property) const {
+  Property* iProperty = this->getProperty(property);
+  if (iProperty) {
+    if (iProperty->isRetained())
+      return Interface::get().getSendingPromise().setNode(*this).setProperty(property).setQos(1).setRetained(true);
+    else
+      return Interface::get().getSendingPromise().setNode(*this).setProperty(property).setQos(1);
+  }
 }
```
euphi: Here a reference to SendingPromise MUST be returned. For now, this result in a crash if you want to set a Property that has not be configured.

Solutions:

1. Change signature of HomieNode::setProperty to `SendingPromise *HomieNode::setProperty(...)` and return NULL --> caller can check for NULL
2. Create error message and `abort()` --> Homie-ESP8266 enforces that you can only send to advertised properties
3. Return a generic SendingPromise& to be able to send on arbitrary topics (old behaviour. Maybe that is what you 

wanted to do and just messed with the nested `if`s?

I suggest to stick to the Homie-2.0 behaviour for now (third solution)

euphi: Just change the line 81 und 82 to one line:
```C++
if (iProperty && iProperty->isRetained())
```

src/HomieNode.hpp
```diff
@@ -33,23 +38,35 @@ class PropertyInterface {
...
-  explicit Property(const char* id, bool range = false, uint16_t lower = 0, uint16_t upper = 0) { _id = strdup(id); _range = range; _lower = lower; _upper = upper; _settable = false; }
+  explicit Property(const char* id) {
+    _id = strdup(id); _name = ""; _unit = ""; _datatype = ""; _format = ""; _retained = true; _settable = false; }
   void settable(const PropertyInputHandler& inputHandler) { _settable = true;  _inputHandler = inputHandler; }
+  void setName(const char* name) { _name = name; }
+  void setUnit(const char* unit) { _unit = unit; }
+  void setDatatype(const char* datatype) { _datatype = datatype; }
+  void setFormat(const char* format) { _format = format; }
+  void setRetained(const bool retained = true) { _retained = retained; }
+
```
euphi: change to `String` (see detailed comment below)

src/HomieNode.hpp
```diff
@@ -33,23 +38,35 @@ class PropertyInterface {
...
-  uint16_t getUpper() const { return _upper; }
   PropertyInputHandler getInputHandler() const { return _inputHandler; }
   const char* _id;
-  bool _range;
-  uint16_t _lower;
-  uint16_t _upper;
+  const char* _name;
+  const char* _unit;
+  const char* _datatype;
+  const char* _format;
```
euphi: these should be stored as String, so it is automatically copied.

Of course most of the Strings that are set here a constants and thus in global memory. 
However, it is possible to receive dynamically created strings here and these could be on stack memory. For example, the allowed values for an ENUM could be generated dynamically.

gorec2005:
@bodiroga Thank for you work - it's work, but ota_update don't :-(

mjcumming:
Is it possible to do a pull on this PR? How?

dikodahan:
+1 on having this released! Even if it will only have partial functionality (like OTA not working yet in initial release).
On the same subject of updating this to Homie 3, the current Homie 2 implementation has a dependency on ArduinoJson 5.x while 6.x has been released for a while now. Would be worth to update that support as well along the way, though back to my first point, could be a later release not to delay the initial support release.

Thank you all for this work, Amazing stuff!!!

euphi:
@mjcumming You can pull  the fork from bodiroga and chekout develop-v3
```
git pull https://github.com/bodiroga/homie-esp8266
cd homie-esp8266
git checkout develop-v3
```

With platformio, you can change your lib_deps from "Homie" to  "https://github.com/bodiroga/homie-esp8266#develop-v3".

jeanvdr:
@bodiroga, gorec2005, dikodahan
for ota_update to work, since topic $online has disappeared from homie v3.0.0, you have to change to 3 lines:
line 16 : client.subscribe("{base_topic}{device_id}/$online".format(**userdata)) > client.subscribe("{base_topic}{device_id}/$state".format(**userdata))
line 69: elif msg.topic.endswith('$online'): > elif msg.topic.endswith('$state'):
line 70: if msg.payload == 'false': > if msg.payload != 'ready':

That's it

mjcumming:
@euphi thank you for the instructions!

euphi:
It seems this branch sometimes looses events:
```
Triggering WIFI_DISCONNECTED event...
↕ Attempting to connect to Wi-Fi...
〽 Sending statistics...
  • Wi-Fi signal quality: 100%
  • Uptime: 1488s
✔ Wi-Fi connected, IP: 192.168.0.145
Triggering WIFI_CONNECTED event...
↕ Attempting to connect to MQTT...
〽 Sending statistics...
  • Wi-Fi signal quality: 8%
  • Uptime: 1548s
〽 Sending statistics...
  • Wi-Fi signal quality: 8%
  • Uptime: 1608s
```

--> MQTT Connected event is missing, although the homie-esp8266 client reconnected to MQTT. The LED is still blinking quickly. However, the client receives set-commands and handles them correctly, so this is a minor issue.

src/SendingPromise.cpp
jeanvdr: @euphi 
How can we do that ?

euphi: Just add after line 57 (`strcat(topic, rangeStr);`):

```C++
  _range.isRange = false;
  _range.index = 0;
```

It is a quick workaround. I think the range-mechanism has to be rewritten.

davidgraeff:
Actually. The over-use of the Arduino String class and std::function will fragment the heap memory very fast and I assume this firmware is just crashing from time to time. The esp is quite fast with rebooting and reconnecting so that may not have been noticed yet.

src/SendingPromise.cpp
jeanvdr: @euphi 
Thanks

euphi:
@davidgraeff RTOS seems to be quite robust.

When using many debug outputs with format strings, I have some problems that  outgoing MQTT messages are lost, but no crashes/resets.

All of my homie devices run very stable (uptime is months and I never saw any reset)

Exception:
* one device using the WS2812FX library resets very often (5-30 minutes). I investigated a lot and haven't found any issue yet, so I more and more think this is a power supply issue.

* some devices with  bad Wifi connections start to reconnect/loose connection over and over, once they lost there wifi-connection for first time.
I think that maybe the initial amount of MQTT traffic is too much for a slow/disturbed Wifi connection.  However, this is not a reset.

* one device has I2C problems every now and then. For now, I blame the I2C device (PCF8575) for that.

euphi:
Initial development for Homie Convention V3.0.x by @bodiroga

PR #561, opened 2019-01-22
Fix documentation URL

luebbe:


luebbe:
This URL has also to be fixed in the title of the project itself: 
"ESP8266 framework for Homie, a lightweight MQTT convention for the IoT..."

PR #563, opened 2019-01-24
Added compiler flag for disabling mDNS

kleini:
Disabling mDNS in homie-esp8266 saves 6456 bytes of flash. I am trying to reduce the overall firmware size by disabling functionality, that I don't need.

I use ESP-01 and the 1MB flash size is too small to get OTA working. Therefore making homie-esp8266 is maybe an option to reduce firmware size and make OTA work on small devices.


euphi:
Merged to #develop and #develop-v3

PR #565, opened 2019-01-29
added compiler flag for disabling the config mode

kleini:
The added compiler flag that allows to remove the config mode from the code completely during compile run. This reduces the size of my empty test firmware by 50760 bytes.


euphi:
Many thanks! After recent arduino / framework updates some simple programs were larger than 434160 byte (limit for 512kb Flash). With the patch, size is much smaller again :-)

I also merged it to develop-v3.

PR #566, opened 2019-02-10
Add an event that is fired whenever statistics are sent 

luebbe:
I felt the need for an event that gets fired whenever Homie sends its statistics. 
E.G. currently I use this to send the free heap via MQTT.


euphi:
Good idea. This is also useful to send sensor data in the the interval as the statistics.

PR #572, opened 2019-03-06
Develop v3 esp32

euphi:
PR with correct target branch (develop-v3)


euphi:
Merged ESP32 implementation from @tantive ! Many thanks!

PR #577, opened 2019-03-10
SSL support for MQTT connection

kleini:
This is stolen from timpur's v2.1 branch.
I like to encrypt my MQTT connections because passwords are transmitted over MQTT in clear text. Passwords transmitted in clear text should be always avoided.

PR #579, opened 2019-03-10
Extended documentation for deep-sleep

kleini:
In my testcases the provided example for deep-sleep mostly caused the properties of some HomieNodes sent in HomieNode::loop() method not to be received by the MQTT broker. Maybe this is due to new async SSL connection handling of my other pull request #577. The introduced postpone of going into deep sleep solved properly the properties of HomieNodes to be received by the MQTT broker.

Maybe there is a better solution by asking the MQTTclient for its state about sending messages, but I did not find one. If you point me to the right place, I will try to fix this correctly.

### DIFF
--- a/docs/advanced-usage/deep-sleep.md
+++ b/docs/advanced-usage/deep-sleep.md
@@ -1,13 +1,20 @@
-Before deep sleeping, you will want to ensure that all messages are sent, including the `$online →  false`. To do that, you can call `Homie.prepareToSleep()`. This will disconnect everything cleanly, so that you can call `ESP.deepSleep()`.
+Before deep sleeping, you will want to ensure that all messages are sent, including property values of your `HomieNode` and the `$online →  false`. To do that, you can call `Homie.prepareToSleep()`. But the event `MQTT_READY` is emitted before your `HomieNode`s `loop()` method is called. Therefore you need to postpone the call to `Homie.prepareToSleep()` until the `loop()` method of all `HomieNode`s is called and their properties are submitted over MQTT. For that the additional Timer is necessary. Then `Homie.prepareToSleep()` will disconnect everything cleanly, so that you can call `Homie.doDeepSleep()`.
 
 ```c++
 #include <Homie.h>
+#include "Timer.h"
+
+Timer t;
+
+void prepareSleep() {
+  Homie.prepareToSleep();
+}
 
 void onHomieEvent(const HomieEvent& event) {
   switch(event.type) {
     case HomieEventType::MQTT_READY:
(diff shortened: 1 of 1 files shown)


stritti:
@kleini I think it would be good if you add part of your PR comments also to the docs to describe for new users why it is done that way.

kleini:
I need to debug once more, what is going on towards deep sleep and why are MQTT messages are not properly sent if I use the original example code.

kleini:
The documentation is improved. But my gut feeling tells me, that a new event is necessary providing information about a successful processing of the loop() method of all HomieNodes. But this can be another merge request.

MoJo2600:
Shouldn't it be `Homie.doDeepSleep()` instead of `ESP.deepSleep()`? The Homie method is not doing much currently, but I think there is a purpose for having a framework method to do deep sleeping.

kleini:
@MoJo2600 you're right. I did not change that sentence and did not even notice, that it is wrong.

kleini:
Fixed and ready to be merged.

docs/advanced-usage/deep-sleep.md
```diff
@@ -1,13 +1,20 @@
-Before deep sleeping, you will want to ensure that all messages are sent, including the `$online →  false`. To do that, you can call `Homie.prepareToSleep()`. This will disconnect everything cleanly, so that you can call `ESP.deepSleep()`.
+Before deep sleeping, you will want to ensure that all messages are sent, including property values of your `HomieNode` and the `$online →  false`. To do that, you can call `Homie.prepareToSleep()`. But the event `MQTT_READY` is emitted before your `HomieNode`s `loop()` method is called. Therefore you need to postpone the call to `Homie.prepareToSleep()` until the `loop()` method of all `HomieNode`s is called and their properties are submitted over MQTT. For that the additional Timer is necessary. Then `Homie.prepareToSleep()` will disconnect everything cleanly, so that you can call Homie.doDeepSleep()`.
```
MoJo2600: I think you missed a  \` here: Homie.doDeepSleep()\`

MoJo2600:
The code is working perfectly. I already implemented it in one of my homie devices. Thank you. I found a typo and left a comment.

docs/advanced-usage/deep-sleep.md
kleini: Sometimes it is too late in the evening... Fixed.

stritti:
one question: merge into `develop` or `develop-v3`? Or both ...

kleini:
> one question: merge into `develop` or `develop-v3`? Or both ...

This documentation is relevant for both.

stritti:
Merged to `develop` and `develop-v3`

PR #582, opened 2019-03-11
Doc corrections: old github-URLs

stritti:
I replaced several places of the old github-links


stritti:
oh, something went wrong ...

PR #580, opened 2019-03-11
[v3] new example: Ping-Node

stritti:
New simple example showing how to create own `HomieNode`s and how to handle MQTT messages.

PR #584, opened 2019-03-12
[WIP] [v3] Initial Migration Guide v2 to v3

stritti:
Uncomplete Migration Guide fo migrate from v2 to v3.

Please comment and let me know what is missing.


kleini:
run loop disconnected is possible with develop branch heading towards v2.1, too. Do we want guides from v2 to v3 only or from v2 to v2.1 and then to v3?

stritti:
@kleini that is a good question. I just collected what I found on using v3.
I haven't used v2.1. IMHO we should describe migration for major releases but could add some notes.

What do you think?

kleini:
I stumbled over run loop disconnected in the develop branch. As I already tried to point out in the Gitter chat, that all current implementations of HomieNodes and even your PingNode example of merge request #580 are broken with the current run loop disconnected implementation on the develop branch and develop-v3 branch.

The loop method of HomieNodes is called once MQTT is connected. `Interface::get().getMqttClient().connected()` is sufficient to run the loop method of HomieNodes. But if within that loop method some MQTT properties should be sent, it fails. `SendingPromise.send()` checks for `Interface::get().ready`. which is not true then.

Therefore even if I leave `runLoopDisconnected` to `false`, I get errors that node properties could not be sent. And I still don't know, whether @euphi intended, that all HomieNodes need to be changed or not?

docs/others/upgrade-guide-from-v2-to-v3.md
```diff
@@ -0,0 +1,15 @@
...
+
+The Homie convention has been revised to v3 to be more extensible and introspectable. Be sure to [check it out](https://github.com/homieiot/convention/tree/v3.0.1).
+
+## API changes in the sketch
+
+1. Constructor of `HomieNode` needs third mandatory param `const char* type`:
+   E.g. `HomieNode lightNode("light", "Light");` -> `HomieNode lightNode("light", "Light", "switch");`.
+2. Signature of handleInput has changed to: `handleInput(const HomieRange& range, const String& property, const String& value)`
+   TODO: see Ping example.
+3. The `loop`s of HomieNodes could be run although there is no MQTT connection by setting param `setRunLoopDisconnected(true);`.
```
euphi: Please note that this change is also in #develop (dev-branch of v2).

stritti: Okay, shall we **move** it then in the v2 migration guide or **add** it there too?

euphi: I think it should be added to a upgrade guide for 2.0 to 2.1 if we ever release a 2.1.

stritti: point `3` is removed

PR #593, opened 2019-04-16
Fix OTA problems for ESP32

### LINKED ISSUE #575: Esp32, OTA, MD5 checksum
As you all probably know already, the Esp32-OTA implementation is not working properly.
There's a problem with the getSketchMD5 esp32 function which ends in a boot loop. This function works properly in other sketchs and if we move the MD5 calculation from where it is (begining of bootnormal.cpp) in the bootnormal::setup procedure, it works without any problems.

bodiroga:
Hi guys!

Thanks to the great instructions provided by @jeanvdr and @bertmelis, OTA updates are now working for ESP32 boards. I don't have any ESP8266 with me right now, so please test it on those boards before merging!

Closes #575

Signed-off-by: Aitor Iturrioz <riturrioz@gmail.com>

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -2,7 +2,6 @@
 
 using namespace HomieInternals;
 
-#ifdef ESP32
 BootNormal::BootNormal()
   : Boot("normal")
   , _mqttReconnectTimer(MQTT_RECONNECT_INITIAL_INTERVAL, MQTT_RECONNECT_MAX_BACKOFF)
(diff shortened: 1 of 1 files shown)


tantive:
@bodiroga Tested OTA on a M5Stick (ESP32), works like a charm. Thank you!
Should be save to merge this PR.

bodiroga:
@stritti, how many approves do we need to merge a PR?

bertmelis:
> @stritti, how many approves do we need to merge a PR?

for a confirmed bugfix I'd hope only one.

stritti:
@bodiroga I hope also one approval should be enough. I am very new in the project and hoped that e.g. @euphi will merge.

stritti:
I merged now.

Issue #351, opened 2017-06-04
Homie 2.0 remains forever waiting for Wi-Fi connection


cecilios:
I have a Homie node in operation (boot normal mode) working ok. If I press the RESET button in the ESP8266 the firmware restarts. After the initial debug messages Homie attempts to connect to WiFi but it remains there forever. This is the last debug message:

↕ Attempting to connect to Wi-Fi...

After several hours of investigation I've discovered that if I remove power supply from ESP8266, wait for a couple of minutes and power it up again, now the node works ok. It looks as if Homie get stacked in a wait loop for the wifi connection event, but this event never takes place because the wifi connection still exists (probably, the time required to reset the system is so small than the WiFi station didn't notice that the ESP8266 was disconnected and so the wifi connection was not dropped. And when the ESP8266 asks to connect to wifi it never receives a new connection as it still exists. Not sure if this is the cause but it is the only explanation I have. But I don't know how to fix this behaviour. 

I also have noticed this problem when the node is in operation and I upload a new firmware (direct upload, not using OTA). After the upload is finished, ESP8266 automatically restarts and the new firmware is executed (boot normal mode). Homie attempts to connect to WiFi but it remains there forever. And the only solution is to remove power supply for a couple of minutes.

I'm using Homie 2.0 from development repo.

bertmelis:
I believe it's a bug in the Arduino framework. Power cycling the device solves the issue. Can't find the issue though...

cecilios:
For now I have bypassed the problem by forcing a wifi disconnection

        `WiFi.disconnect();`

in my node setup when not in config mode

marvinroger:
I'll take a look!

johnjore:
For what it's worth, I've seen the same issue when working on a reed sensor, frequent connect / disconnect. The code here didn't have the same problem, https://github.com/8n1/ESP8266-Tiny-Door-and-Window-Sensor.

andreasbrinch:
I have the same issue, but only when I have a static ip configured. It cant connect after a reset (using the button on wemos) or after deep sleep. Everything works when the static ip is not configured and after a power off.

andreasbrinch:
And I dont think it is related to the time the device is off. I can repower the device with only a second in delay which works and it doesn't with a reset of 2-3 seconds.

highno:
I can confirm this bug (tho I can't exactly see, what the sketches changes were that led to this bug).
[...]

PR #598, opened 2019-04-26
add workaround for issue #351

stritti:
Workaround for issue #351

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -161,6 +161,8 @@ void HomieClass::setup() {
     return;  // never reached, here for clarity
   }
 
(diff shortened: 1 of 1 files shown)


stritti:
could/should also be merged in `develop` for V2.0

bodiroga:
Here we can see more reports about ESP8266s having problems reconnecting to the WiFi network: https://github.com/esp8266/Arduino/issues/5527, but I'm not sure how we should proceed with all these .persistent(false), .disconnect(), .mode(WIFI_STA),... 😞

stritti:
Okay I'll merge as workaround, but leave issue open for further investigations to solve "cleaner".

PR #600, opened 2019-05-02
Added deep sleep support for ESP32

EinfachArne:
I added support for deep sleep with an ESP32.

Changes I made:

- HomieClass::prepareToSleep() is now used for both platforms (ESP8266 & ESP32) but remains unchanged.

- HomieClass::doDeepSleep() is new for the ESP32. The only change compared to the 8266 version is that the method for ESP32 is called: esp_deep_sleep_start();

With this solution the ESP will never wake up, unless you define a way to wake it up for yourself. I can change this so that it automatically wakes up after a defined period of time, just like with the 8266. But I think the user should be able to choose his own wake up method as described here: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/sleep_modes.html

My question is: Should there be a way within the framework to choose the preferred way or should the user do it by himself, outside the framework?


kleini:
Good, that the linter already told you, where you have to place the opening curly braces.

EinfachArne:
Thanks for your input. Now I adapted the API to be almost the same as for the 8266.

Sorry for the formatting changes, I think I reverted all of them. At least the linter is happy with it :)

andreacioni:
Hi guys! It will be awesome to allow ESP32 users to take advantage of the wake up capabilities that the hardware offers. Something like this should work:
```cpp
void HomieClass::doDeepSleep(uint32_t time_us = 0, gpio_num_t gpio_pin=-1, int logic_level=-1) {
  Interface::get().getLogger() << F("💤 Device is deep sleeping...") << endl;
  Serial.flush();

  if(gpio_pin > 0 && logic_level > 0) {
    esp_sleep_enable_ext0_wakeup(gpio_pin,logic_level);
  }

  esp_sleep_enable_timer_wakeup(time_us);
  esp_deep_sleep_start();
}
```

What do you think?

EinfachArne:
Perfectly fine for me but I am not the one to decide this. I added your suggestion into my pull request. Now let us see what the maintainers decide.

kleini:
@EinfachArne you should rebase your branch on develop-v3 instead of merging develop-v3 changes into your branch. This will result in a much cleaner Git history.

stritti:
@EinfachArne & @kleini could you rebase to branch `develop` please?

Issue #323, opened 2017-03-27
Config does not contain a settings object


Gulaschcowboy:
Hi @marvinroger et al.

I have defined 4 custom settings and I can't update them via MQTT nor via App/Web. To narrow down the issue I tried same with your CustomSettings example. If I don't change it and keep just 1 custom setting I can update the settings with :
 mosquitto_pub -h broker -t "homie/test-3/\$implementation/config/set" -u user -P pass -m '{"settings":{"temperatureInterval":30}}' -r

As soon as I add my settings it fails. After publishing with the above command I get
 ✖ Config does not contain a settings object
 ✖ Configuration not updated
in serial monitor.

See my modified CustomSettings example attached

[test-settings.ino.txt](https://github.com/marvinroger/homie-esp8266/files/873696/test-settings.ino.txt)

**Update:**
I could narrow it down further. If I don't set more than 1 value as default, all values become settable.

Thank you
Alex

exxamalte:
Just looking at the code ([Config.cpp](https://github.com/marvinroger/homie-esp8266/blob/develop/src/Homie/Config.cpp), lines 246-263), it appears that the error message "Config does not contain a settings object" is thrown when trying to set config values but `settings` is not defined yet, i.e. had not been defined before.
Does the config file on your device contain a `settings` key initially?

Gulaschcowboy:
@exxamalte "Does the config file on your device contain a settings key initially?" Yes, this will be the case, but it is just the resulting error. 
As I stated in my Update: If you define more than one HomieSetting with a default value via `setDefaultValue` the error occurs. To reproduce: Define more than one HomieSetting, make more than one defined with a default value using `setDefaultValue.` Set values in the App/Web different from the defaults and those manual, non-default values are not used during runtime and are not settable via MQTT $configuration/set. 

Don't use `setDefaultValue` or only once and they become settable via APP/Web and MQTT. So my uneducated guess is, that there is a parsing error in /Utils/Validation.cpp

kartom:
I can confirm that custom settings does not work when the default values are set. It works fine when not using the default values. 
It would also be good if the custom settings of type double would accept integer values. Now one have to enter a decimal number, but i.e. 12 is in my point of view a perfectly valid double.

marvinroger:
I can't reproduce. Sample sketch:

```c++
#include <Homie.h>

HomieNode temperatureNode("temperature", "temperature");

HomieSetting<long> aSetting("a", "A");
[...]

PR #606, opened 2019-06-10
Incremental config for HomieSettings

MajorTwip:
Incremental config for HomieSettings now working, even when "settings"-Object didn't exist in JSON. closes homieiot/homie-esp8266#605


MajorTwip:
binary apparently is 5kb too big for the esp01... I don't think this is due to my 2 sloc...

MajorTwip:
Would address #323 too

MajorTwip:
Not suitable anymore due to merges to migrate to Arduinojson 6.
Will reevaluate and if needed reopen.

Issue #471, opened 2018-01-07
Size of binary grows by 15-20% with latest version


luebbe:
Hi folks,

maybe not the 100% correct place to ask the question here, but I'm hoping for some tips nevertheless.

The binary size of my https://github.com/luebbe/homie-sonoff/ project grew by 20% (from 366544 bytes to 416192 bytes) by switching to the latest version of homie and updating all the dependencies. This is becoming a bit of a problem on sonoff devices which only have 1 MByte of flash.
I don't really understand what is causing this.
Is it possible that the linker just links everything, even unused code from piolibdeps into the binary? 
How can I prevent this from happening?

timpur:
im going to be looking into this also with v2.1

luebbe:
FYI I made a little experiment with my homie-ota project. In https://github.com/luebbe/homie-ota/tree/playground. I changed platformio.ini so that four builds are made in the following order

- serial1 with no "extra" lib_deps
- ssd1306 with the ssd1306 dependency
- u8g2 with the u8g2 dependency
- serial2 with no "extra" lib_deps (just a clone of serial1)

if you delete .piolibdeps before "pio run", you can see that the ssd1306 and u8g2 lib_deps are downloaded and installed when needed.
The resulting binary size of all four builds is exactly the same (404KByte), which I find mildly confusing. I would have expected more for the "ssd1306" and "u8g2" builds.
Or do the binaries grow in fixed chunks, because only multiples of n KByte can be flashed?

timpur:
Good question, I have no idea but when I have time I hope to find out my self. Will pass on anything I learn.

timpur:
Good question, I have no idea but when I have time I hope to find out my self. Will pass on anything I learn.

timpur:
Good question, I have no idea but when I have time I hope to find out my self. Will pass on anything I learn.

kylegordon:
Can confirm, using https://github.com/euphi/Homie_BareMinimum results in a 395K binary, with Homie v2.0.0-beta.3, v2.0.0-beta.2, and Develop

This also allows a one time flash of the Sonoff device I'm using at the moment ( Sonoff Touch with ESP82858 ), and then all future OTA updates fail with an error 400 NOT_ENOUGH_SPACE

Has Homie gotten fat? :-) ( I don't think so, I think it's likely to be a dependency somewhere )

timpur:
[...]

PR #611, opened 2019-07-15
Fixed device not reconnecting after MQTT connection loss

dakhnod:
MQTT was not reconnecting after a loss, since BootNormal::_onMqttDisconnected had to be called [twice](https://github.com/homieiot/homie-esp8266/blob/develop/src/Homie/Boot/BootNormal.cpp#L500) in order to periodically try to reconnect.

In my tests it got called only once, leading to only one singe reconnection attempt.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -518,9 +518,9 @@ void BootNormal::_onMqttDisconnected(AsyncMqttClientDisconnectReason reason) {
 
     _mqttConnect();
 
-  } else {
-    _mqttReconnectTimer.activate();
   }
+  _mqttReconnectTimer.activate();
+  
 }
 
 void BootNormal::_onMqttMessage(char* topic, char* payload, AsyncMqttClientMessageProperties properties, size_t len, size_t index, size_t total) {
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -517,10 +517,8 @@ void BootNormal::_onMqttDisconnected(AsyncMqttClientDisconnectReason reason) {
     }
 
     _mqttConnect();
-
   }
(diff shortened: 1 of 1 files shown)


kleini:
Tests need to be fixed first.

dakhnod:
"Error: The program size (435040 bytes) is greater than maximum allowed (434160 bytes)"

i don't really think there is much i can do about that

kleini:
I tried to fix the tests for pull request #577 .

dakhnod:
please help me explain:

i literally deleted three lines. how can that affect the outcome of build teste in that way?

bertmelis:
I would let pass. It's the esp01 which lacks memory.

I don't know how CircleCI handles things, but for esp01 one should test without config or mdns enabled to lower memory usage.

kleini:
@bertmelis I switched from esp01 to esp01_1m in pull request #577. This increases ROM from 512K to 1M and compiles do work again. We can then discuss somewhere else, whether we want to stay with esp01_1m or disable config mode for esp01.

kleini:
> please help me understand:
> 
> i literally deleted three lines. how can that affect the outcome of build teste in that way?

The resulting image size dynamically depends on the used libraries and their versions. Especially the Arduino framework seems to grow with every release. Please see issue #471 for the according discussion.

dakhnod:
> I did not test the code but it looks reasonable.

I am sure the code used to work once, i just don't get why onMqttDisconnected() should be called twice.

Is there any scenario where that would happen after a disconnect?

kleini:
I will try to test that by blocking the MQTT port of my MQTT server, so I can bring the devices into the problem, that a second connection retry becomes necessary.

dakhnod:
Jup, exactly what I did

stritti:
I think this is also issue in `develop-v3`.

stritti:
I cherry picked it to branch `develop-v3`

PR #616, opened 2019-07-23
ArduinoJson6 migration

iluminat23:
Hi,

I'm working on the ArduinoJson6 support. I did compile time tests and a short test with my temperature sensor.

### DIFF
--- a/src/Homie/Boot/BootConfig.cpp
+++ b/src/Homie/Boot/BootConfig.cpp
@@ -121,14 +121,15 @@ void BootConfig::loop() {
 
 void BootConfig::_onWifiConnectRequest(AsyncWebServerRequest *request) {
   Interface::get().getLogger() << F("Received Wi-Fi connect request") << endl;
-  DynamicJsonBuffer parseJsonBuffer(JSON_OBJECT_SIZE(2));
+  DynamicJsonDocument parseJsonDocument(JSON_OBJECT_SIZE(2));
   const char* body = (const char*)(request->_tempObject);
-  JsonObject& parsedJson = parseJsonBuffer.parseObject(body);
-  if (!parsedJson.success()) {
+  auto error = deserializeJson(parseJsonDocument, body);
+  if (error) {
     __SendJSONError(request, F("✖ Invalid or too big JSON"));
     return;
   }
 
+  JsonObject parsedJson = parseJsonDocument.as<JsonObject>();
   if (!parsedJson.containsKey("ssid") || !parsedJson["ssid"].is<const char*>() || !parsedJson.containsKey("password") || !parsedJson["password"].is<const char*>()) {
     __SendJSONError(request, F("✖ SSID and password required"));
     return;
@@ -143,8 +144,8 @@ void BootConfig::_onWifiConnectRequest(AsyncWebServerRequest *request) {
 void BootConfig::_onWifiStatusRequest(AsyncWebServerRequest *request) {
   Interface::get().getLogger() << F("Received Wi-Fi status request") << endl;
 
-  DynamicJsonBuffer generatedJsonBuffer(JSON_OBJECT_SIZE(2));
-  JsonObject& json = generatedJsonBuffer.createObject();
+  DynamicJsonDocument generatedJsonDocument(JSON_OBJECT_SIZE(2));
+  JsonObject json = generatedJsonDocument.as<JsonObject>();
   String status;
 
   //String json = "";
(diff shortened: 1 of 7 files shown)


kleini:
Impressive work. Will try to review and test that.

src/Homie/Boot/BootConfig.cpp
```diff
@@ -339,25 +341,25 @@ void BootConfig::_onDeviceInfoRequest(AsyncWebServerRequest *request) {
...
     jsonNode["id"] = iNode->getId();
     jsonNode["type"] = iNode->getType();
     nodes.add(jsonNode);
   }
 
-  JsonArray& settings = json.createNestedArray("settings");
+  JsonArray settings = json.createNestedArray("settings");
   for (IHomieSetting* iSetting : IHomieSetting::settings) {
-    JsonObject& jsonSetting = jsonBuffer.createObject();
+    JsonObject jsonSetting = jsonDocument.as<JsonObject>();
```
codefrey: With `jsonDocument` not initialized this will lead to undefined behaviour as described above. With `jsonDocument` initialized, this will put the attributes of the last setting into the root of `jsonDocument`.
Use `JsonObject jsonSetting = settings.createNestedObject();` instead and remove the add command below.

MajorTwip:
Compiles fine on NodeMCU0.9.
OTA works fine.

Configuration Update not working:

✖ Invalid or too big JSON
✖ Configuration not updated

with following content:
```
{
    "name": "CSHPower"
}
```
which works fine with develop-v3


Update sent with https://github.com/MajorTwip/HomieControl

src/Homie/Boot/BootConfig.cpp
```diff
@@ -339,25 +341,25 @@ void BootConfig::_onDeviceInfoRequest(AsyncWebServerRequest *request) {
...
-  JsonObject& firmware = json.createNestedObject("firmware");
+  JsonObject firmware = json.createNestedObject("firmware");
   firmware["name"] = Interface::get().firmware.name;
   firmware["version"] = Interface::get().firmware.version;
 
-  JsonArray& nodes = json.createNestedArray("nodes");
+  JsonArray nodes = json.createNestedArray("nodes");
   for (HomieNode* iNode : HomieNode::nodes) {
-    JsonObject& jsonNode = jsonBuffer.createObject();
+    JsonObject jsonNode = jsonDocument.as<JsonObject>();
```
codefrey: With `jsonDocument` not initialized this will lead to undefined behaviour as described above. With `jsonDocument` initialized, this will put the attributes of the last node into the root of `jsonDocument`.
Use `JsonObject jsonNode = settings.createNestedObject();` instead and remove the add command below.

src/Homie/Boot/BootConfig.cpp
```diff
@@ -143,8 +144,8 @@ void BootConfig::_onWifiConnectRequest(AsyncWebServerRequest *request) {
 void BootConfig::_onWifiStatusRequest(AsyncWebServerRequest *request) {
   Interface::get().getLogger() << F("Received Wi-Fi status request") << endl;
 
-  DynamicJsonBuffer generatedJsonBuffer(JSON_OBJECT_SIZE(2));
-  JsonObject& json = generatedJsonBuffer.createObject();
+  DynamicJsonDocument generatedJsonDocument(JSON_OBJECT_SIZE(2));
+  JsonObject json = generatedJsonDocument.as<JsonObject>();
```
codefrey: Undefined behaviour as described below.

src/Homie/Boot/BootConfig.cpp
```diff
@@ -175,20 +176,21 @@ void BootConfig::_onWifiStatusRequest(AsyncWebServerRequest *request) {
 
   json["status"] = status;
   String output;
-  json.printTo(output);
+  serializeJson(json, output);
```
codefrey: `serializeJson(generatedJsonDocument, output);` can be used here when the issue above is resolved.

codefrey:
If you want, you can check out my develop-v3 branch. I did the same things but in another way.

iluminat23:
Closing the pull request to make fixes.

codefrey:
@iluminat23 If you don't mind I would like you to consider my implementation of the migration (#618) and [cleanup code](https://github.com/codefrey/homie-esp8266/commit/2009746c5e3dd8a5a453fd0e451def4520bd594b) instead.

I've already based a WPA2 enterprise implementation for ESP32 off of if, that I would like to upstream later and it would reduce my work to adapt the code.

iluminat23:
@codefrey I'm fine with the migration patch from your branch. I will test it today. But the "cleanup code" is a massive patch. The commit message doesn't fully cover what is done. I'm not a maintainer here just someone who wanted to contribute as you. But I think the cleanup patch should be split in smaller patches. This makes the review easier.

Also you are using some kind of Hungarian notation which is used no where else in this code. But this is up to the maintainers of this code base.

codefrey:
@iluminat23 Thanks for your understanding. What do you mean by Hungarian notation? Do you mean the scoping of the instructions for separate values in the validation code? I did this to logically group the validation items and to prevent coding errors from occuring by having too many JsonVariant objects in one scope.

And I can split the cleanup code by files and change the commit message if you like. Do you have any suggestions for a more descriptive message? I noticed I've included a change of the patching algorithm too, so I will make a separate commit out of it.

iluminat23:
@codefrey
It's about the prefixing variable names with their types (https://en.wikipedia.org/wiki/Hungarian_notation).

For example this one `JsonVariant jvWiFiPassword` The `jv` prefix is making the code harder to read. And as this style is not used in the code base it looks alien. I'm a fan of short variable names so in this case I would go with `password` or maybe `wifiPassword`.

But as I said I'm not a maintainer of this code base. If the maintainers are fine with your codingstyle I will not judge you.

codefrey:
I think I started doing this prefixing because in Config::load() there would have been a naming conflict between the JsonVariants and the variables their values would be written to if they existed in the config.

I will change it and add a `Var` suffix to the JsonVariants in Config::load() instead. Thank you for your suggestions!

PR #618, opened 2019-08-20
Migrate to ArduinoJson version 6

codefrey:
Migrates the code of develop-v3 from ArduinoJson v5 to v6.

I tested this patch on ESP32 and haven't encountered any issues except the [JSON example configuration](https://github.com/homieiot/homie-esp8266/blob/develop-v3/docs/configuration/json-configuration-file.md) not being able to be loaded because it is too big as it is duplicated by ArduinoJson.

I'd recommend someone to test this patch / the branch on my repo on ESP8266 first before merging since I don't have one available.

### DIFF
--- a/library.json
+++ b/library.json
@@ -22,7 +22,7 @@
   "dependencies": [
     {
       "name": "ArduinoJson",
-      "version": "^5.10.0"
+      "version": "^6.11.4"
     },
     {
       "name": "AsyncMqttClient",
--- a/src/Homie/Boot/BootConfig.cpp
+++ b/src/Homie/Boot/BootConfig.cpp
@@ -129,14 +129,16 @@ void BootConfig::_onWifiConnectRequest(AsyncWebServerRequest *request) {
   }
 
   JsonObject parsedJson = parseJsonDoc.as<JsonObject>();
+  JsonVariant jvWiFiSsid = parsedJson["ssid"];
+  JsonVariant jvWiFiPassword = parsedJson["password"];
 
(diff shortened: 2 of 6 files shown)


stritti:
Dublicate: https://github.com/homieiot/homie-esp8266/pull/616

codefrey:
Maybe I should not only have looked in issues...

Should I close this and help reviewing the code of the duplicate request? I did a few things differently, maybe you can take a look at it.

stritti:
> Should I close this and help reviewing the code of the duplicate request? I did a few things differently, maybe you can take a look at it.

That would be awesome. Many thanks for your support

iluminat23:
I've tested your migration code and it breaks the configurator. After putting in wifi and mqtt data the configurator hangs.

Tested with an esp8266 and https://homieiot.github.io/homie-esp8266/configurators/v2/

codefrey:
I looks like the problem is caused by the missing `settings` object in the JSON received by GET on `/device-info`. Which is caused by the `jsonDoc` not having enough initial space to create the array for it, because the root object size is not included in the size calculation.

But I have no clue why this has noot been a problem prior to my commit.

codefrey:
There appear to be more occurences of this "coding error" in the WiFi status, connect and proxy control requests which also don't work. I entered an example JSON into the [ArduinoJson Assistant](https://arduinojson.org/v6/assistant/), but it gave me the size calculations that are currently used. Maybe this is a coding error on my part, but I am unable to spot it.

codefrey:
Closing this pull request in favor of #622. I've reworked the basic migration code to resolve the memory related issues from the migration there.

I will create separate pull requests for the other code optimizations related to ArduinoJson after #622 has been tested and merged.

Issue #588, opened 2019-03-20
[develop-v3] Core panics after connection to AP


EinfachArne:
Hey,

everything works fine until I connect to my ESP32 via WiFi. After that I keep getting the following error message in an endless loop:

```
✖ /homie/config.json doesn't exist
Configuration invalid. Using CONFIG MODE
💡 Firmware bare-minimum (1.0.0)
🔌 Booting into config mode 🔌
Device ID is 30aea41f1b18
AP started as Homie-30aea41f1b18 with IP 192.168.123.1
Triggering Wi-Fi scan...
dhcps: send_offer>>udp_sendto result 0
Guru Meditation Error: Core  0 panic'ed (InstrFetchProhibited). Exception was unhandled.
Core 0 register dump:
PC      : 0x00000000  PS      : 0x00060030  A0      : 0x8010ba31  A1      : 0x3ffb4e20
A2      : 0x3ffd03a4  A3      : 0x3ffcfde0  A4      : 0x3ffce900  A5      : 0x3ffce654
A6      : 0x027ba8c0  A7      : 0x0c7ba8c0  A8      : 0x8010b8d0  A9      : 0x3ffb4de0
A10     : 0x3ffd03b4  A11     : 0x3ffcfde0  A12     : 0x3ffb4e2c  A13     : 0x00000044
A14     : 0x00000001  A15     : 0x00000006  SAR     : 0x00000010  EXCCAUSE: 0x00000014
EXCVADDR: 0x00000000  LBEG    : 0x4000c349  LEND    : 0x4000c36b  LCOUNT  : 0x00000000

ELF file SHA256: 0000000000000000000000000000000000000000000000000000000000000000

Backtrace: 0x00000000:0x3ffb4e20 0x4010ba2e:0x3ffb4e60 0x40101e21:0x3ffb4e80 0x401057f5:0x3ffb4ec0 0x4010aa72:0x3ffb4ee0 0x400fa53b:0x3ffb4f00 0x4009180d:0x3ffb4f30

Rebooting...
```

After I disconnect the core stops to panic but that makes it impossible to upload a config.

I used the current development version of Homie-V3 on a Wemos Lolin32 V1.0.0.

Any ideas what the cause could be?

bertmelis:
Could you decode the backtrace?

euphi:
I don't use config mode, so I did not tested it yet. 

What happesn, if you upload a `homie/config.json` to SPIFFS?

[...]

PR #619, opened 2019-08-21
Add workaround for AP crash issues on ESP32

codefrey:
This is a workaround for issue #588. Disabling WiFi persistence will stop the crashes from occuring, as long as no other firmware with WiFi persistence has been ran on the ESP32 prior. In this case the internal flash needs to be wiped beforehand.


kleini:
The same seems to apply for ESP8266. I need to wipe the complete flash before using it with homie-esp8266 otherwise strange issues occur. I have to read about WiFi persistence. This little chip has too many strange functions doing strange little things.

PR #628, opened 2019-10-15
Use recursive config patching

codefrey:
* More flexible patching
* Better readable code

### DIFF
--- a/src/Homie/Config.cpp
+++ b/src/Homie/Config.cpp
@@ -304,26 +304,7 @@ bool Config::patch(const char* patch) {
   deserializeJson(configJsonDoc, configJson);
   JsonObject configObject = configJsonDoc.as<JsonObject>();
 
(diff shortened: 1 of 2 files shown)


codefrey:
This PR will make the config patching algorithm recursive.
This results in the following benefits:
- The patching code is easier to read
- The storage footprint is lower
- All future changes to the config structures and config depth should be covered

The patching has the following behaviour:
- If an element in the patch does not exist in the current config, it will be set to the patch element.
- If an element in the patch is not an object or the element in the current config is not an object, it well be set to the patch element.
- If an element in the patch is an object and the element in the current config is an object all described steps will be repated with these rules.

codefrey:
Please don't merge yet. With the password being optional after merging #647, I can rewrite the patching to support the removal of configuration values by setting them to `null` in the patch.

stritti:
@codefrey could you rebase to branch `develop` please?

PR #633, opened 2019-10-21
Fix potential MQTT topic buffer overflow

codefrey:
* Fix wrong MQTT topic buffer size calulation which might lead to a buffer overflow


stritti:
Thank you very much for your improvement.

Issue #641, opened 2019-11-24
Commit 91e9b63 on develop-v3 causes crash on ESP32


nemidiy:
Hi guys, the last commit pushed to develop-v3 makes ESP32 crash. It does not crash on ESP2866 I tested it as well.

```log
commit 91e9b63c9618b3809a51ee7b0e0c057ad3bda62b (HEAD -> develop-v3, origin/develop-v3)
Author: peret2000 <19463568+peret2000@users.noreply.github.com>
Date:   Tue Nov 12 20:52:24 2019 +0100

    Subscribing to messages using Homie.getMqttClient() is not reliable: topic is truncated (#526)
    
    * Declare _mqttTopicCopy in BootNormal.hpp
    * Using _mqttTopicCopy to manage subscribed topics
    * Removed leading spaces
```

Backtrace :
```log
0x4008c800: invoke_abort at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/esp32/panic.c line 155
0x4008ca31: abort at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/esp32/panic.c line 170
0x40088d1f: xQueueTakeMutexRecursive at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/freertos/queue.c line 621
0x400e5af8: spi_flash_op_lock at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/spi_flash/cache_utils.c line 56
0x400829ea: spi_flash_disable_interrupts_caches_and_other_cpu at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/spi_flash/cache_utils.c line 95
0x40081b23: spi_flash_protected_read_mmu_entry at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/spi_flash/flash_mmap.c line 299
0x400e5628: spi_flash_cache2phys at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/spi_flash/flash_mmap.c line 426
0x400e8366: esp_ota_get_running_partition at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/app_update/esp_ota_ops.c line 508
0x400e5def: do_global_ctors at /Users/ficeto/Desktop/ESP32/ESP32/esp-idf-public/components/esp32/cpu_start.c line 444
[...]

PR #642, opened 2019-11-27
getSketchMD5() at BootNormal constructor crashes on ESP32 (#641)

nemidiy:
Removed the lines copying the sketch MD5 at the BootNormal constructor
since that crashed the firmware. At that point the ESP object is not ready
to invoke getSketchMD5(). Also the code was duplicated since it's was
already invoked at the BootNormal::setup function with no issues.


nemidiy:
Hi @bodiroga 
Yes, I tested it on an ESP32 devkit. Thanks!

Issue #644, opened 2019-12-03
[develop-v3] OTA failing


nemidiy:
Hi people,

I've been hitting my head against the wall trying to OTA update an ESP8266. The code is minimalistic 

```c++
#include "Arduino.h"
#include <Homie.h>


HomieNode sensorNode("sensor", "Sensor", "sensor");

unsigned long lastSent = 0;

void loopHandler() {
  unsigned long m = millis();
  if ( m - lastSent >= 10 * 1000UL || lastSent == 0) {
    Homie.getLogger() << "Fake value: " << m << endl;
    sensorNode.setProperty("value").send(String(m));
    lastSent = millis();
  }
}

void setup() {
  Serial.begin(115200);
  Homie.getLogger() << "kkkkkkkkkkkkkkkkkkkkkk" << endl;
  Homie.getMqttClient().setKeepAlive(120);
  Homie_setFirmware("test", "1.0.0");
  Homie.setLoopFunction(loopHandler);
  sensorNode.advertise("value").setName("Value").setDatatype("double");
  Homie.setup();
}

void loop(){
  Homie.loop();
}
```
The platformio.ini file is :
```ini
[env:nodemcuv2]
platform = espressif8266
board = nodemcuv2
framework = arduino
build_flags =
    -D HOMIE_MDNS=0
    -D PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY
upload_speed = 115200
lib_ldf_mode = deep
lib_deps =
    ArduinoJson@6.13.0
    https://github.com/homieiot/homie-esp8266.git#develop-v3
```
I also repartitioned the mcu to make sure there is enough space:

```
# Name,   Type, SubType, Offset,  Size, Flags
nvs,      data, nvs,     0x9000,  0x6000,
phy_init, data, phy,     0xf000,  0x1000,
factory,  app,  factory, 0x10000, 3M,
```
 Here are my libs :
```
Dependency Graph
|-- <ArduinoJson> 6.13.0
|-- <Homie> 3.0.0 #7818c73
|   |-- <ArduinoJson> 6.13.0
|   |-- <AsyncMqttClient> 0.8.2
|   |   |-- <ESPAsyncTCP> 1.2.2
|   |   |   |-- <ESP8266WiFi> 1.0
|   |-- <Bounce2> 2.52
|   |-- <ESP Async WebServer> 1.2.3
|   |   |-- <ESPAsyncTCP> 1.2.2
[...]

PR #645, opened 2019-12-12
Fix topic copying for chunked messages

codefrey:
Only copy the topic on the first message chunk. Otherwise the old copy will be
destroyed on subsequent calls to `_onMqttMessage()` for the same message, while
the copy is still being referenced by `_mqttTopicLevels`. This leads to undefined
behaviour.

Should solve #644 

Thanks @nemidiy for investigating this issue!

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -781,13 +781,14 @@ void BootNormal::_onMqttDisconnected(AsyncMqttClientDisconnectReason reason) {
 void BootNormal::_onMqttMessage(char* topic, char* payload, AsyncMqttClientMessageProperties properties, size_t len, size_t index, size_t total) {
   if (total == 0) return;  // no empty message possible
 
-  size_t topicLength = strlen(topic);
-  _mqttTopicCopy = std::unique_ptr<char[]>(new char[topicLength+1]);
-  memcpy(_mqttTopicCopy.get(), topic, topicLength);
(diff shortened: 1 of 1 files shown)


kleini:
I tested this pull request on top of pull request #646. OTA partly seems to work. I see some firmware file transferred to my esp8266 devices, which is already a lot more compared to what was possible in the past, but after the firmware upload to the device I get "400 BAD_FIRMWARE".

nemidiy:
I tested the @codefrey's PR on nodemcuv2 and nodemcu32 and works for both platforms.
The only difference is that for esp8266 the OTA updater script says that it failed (even though it did not) and for esp32 says it's successful, don't recall where but someone reported this same thing. Probably some bug in the python script. 

@kleini when you applied the DNS commit as well and tested OTA, were us using dns1 and/or dns2 or did you just happen to apply the commit in #646 and tested without setting dns ? I am asking so that I can try it out myself. Thanks.

kleini:
@nemidiy I tested on Saturday the exactly the following version https://github.com/kleini/homie-esp8266.git#35419c27ada5af4f1fc44fb19f339b355149599f, which is #646 and this pull request on top.

nemidiy:
Hi @kleini I pulled develop-v3 and then merged this on top. I tested both setting up DNS and without DNS. OTA worked for me in both cases (tested on nodemcuv2). maybe you can put the details of your set up in the OTA ticket and I can try and help you out ? thanks.

PR #647, opened 2019-12-19
Various code optimizations

codefrey:
- In the migration to ArduinoJson v6 I've introduced a bug disallowing the usage of `null` as `wifi.password`. Commit 25f6b8c fixes this.
- The parameter `ota.enabled` was treated as optional in the config load code, but not in the validation code. Commit 64ca0ba will make `ota.enabled` optional and default to disabled OTA.
- In Config::getSafeConfigFile(), a `char[]` buffer gets allocated within a `unique_ptr`, only to be copied and and then discarded at the end of the method. Commit 60ca263  removes this unnecessary step.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -472,7 +472,7 @@ void BootNormal::_advertise() {
     {
       char* safeConfigFile = Interface::get().getConfig().getSafeConfigFile();
       packetId = Interface::get().getMqttClient().publish(_prefixMqttTopic(PSTR("/$implementation/config")), 1, true, safeConfigFile);
-      free(safeConfigFile);
+      delete safeConfigFile;
       if (packetId != 0) _advertisementProgress.globalStep = AdvertisementProgress::GlobalStep::PUB_IMPLEMENTATION_VERSION;
       break;
     }
--- a/src/Homie/Config.cpp
+++ b/src/Homie/Config.cpp
@@ -68,7 +68,6 @@ bool Config::load() {
 
   const char* reqName = parsedJson["name"];
   const char* reqWifiSsid = reqWifi["ssid"];
-  const char* reqWifiPassword = reqWifi["password"];
   const char* reqMqttHost = reqMqtt["host"];
 
   /* Optional config items */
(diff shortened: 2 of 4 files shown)


codefrey:
Force pushed the last commit to replace `free(safeConfigFile)` with `delete safeConfigFile`, since the allocation was done using `new`.

src/Homie/Boot/BootNormal.cpp
```diff
@@ -472,7 +472,7 @@ void BootNormal::_advertise() {
     {
       char* safeConfigFile = Interface::get().getConfig().getSafeConfigFile();
       packetId = Interface::get().getMqttClient().publish(_prefixMqttTopic(PSTR("/$implementation/config")), 1, true, safeConfigFile);
-      free(safeConfigFile);
+      delete safeConfigFile;
```
bodiroga: Can anyone else comment on this? My C++ skills are not good enough to review this specific change, sorry.

bodiroga: Ok, understood. The safeConfigFile is created in https://github.com/homieiot/homie-esp8266/blob/60ca263c1c39051bc62dd4e96942067062f4fd3e/src/Homie/Config.cpp#L167

and 'delete' is the correct method to release memory when objects are created by the "new" expression (https://www.geeksforgeeks.org/delete-in-c/).

LGTM!

stritti:
LGTM

src/Homie/Config.cpp
```diff
@@ -164,9 +164,9 @@ char* Config::getSafeConfigFile() const {
   parsedJson["mqtt"].as<JsonObject>().remove("password");
 
   size_t jsonBufferLength = measureJson(jsonDoc) + 1;
-  std::unique_ptr<char[]> jsonString(new char[jsonBufferLength]);
-  serializeJson(jsonDoc, jsonString.get(), jsonBufferLength);
-  return strdup(jsonString.get());
+  char* jsonString = new char[jsonBufferLength];
+  serializeJson(jsonDoc, jsonString, jsonBufferLength);
+  return jsonString;
```
davidgraeff: It's better than before. But you should never return an allocated piece of memory as raw pointer in modern C++.

codefrey:
@davidgraeff I thought about using the String class, but String does not implement move semantics so I kept the pointer. It might make sense to replace it with an unique_ptr though.

PR #649, opened 2020-01-02
Don't run loop if connected to Wifi, but not yet connected to MQTT.

euphi:
Instead, wait until all advertisement has been done.

This is a quick fix - and probably not the cleanest solution.

Explanation:

Due to my change (a year ago or so), to be able to run HomieNode::loop() also if disconnected, the `loop()` function is now also called (even if `rRunLoopDiscconected` is set to false) when Wifi is connected, but not yet MQTT.


stritti:
@euphi could you rebase to branch `develop` please? 

euphi:
See 3efc0ca

PR #650, opened 2020-01-03
added dependencies for ldf mode off

kleini:


bodiroga:
Hi @kleini!

The CircleCI built fails because the library manager doesn't find the 1.0 version of the Ticker library. Can we use the 3.1.5 version of the library?

Thanks!

kleini:
I had similar problems locally. PlatformIO is not able to define proper dependencies for libraries in Arduino core if they are with a same name in PlatformIO library registry. Version 3.1.5 has a very different API and I even doubt, that Ticker in Arduino core is same/similar to Ticker in PlatformIO library registry. I will have a look, whether it makes sense to use Ticker 3.1.5 and therefore I have to look into the implementations, how good or bad implemented they are.

bodiroga:
Hi Marcus! (@kleini)

Thanks for your fast response, but I have one question: if Ticker is already embedded in the esp8266 and esp32 arduino core frameworks, why do we need to declare it in the library.json file? It is going to be always available, isn't it? There must be something that I don't fully understand :disappointed: 

Thank you for your work!

kleini:
Try to build the simplest possible Homie device and configure `lib_ldf_mode = off`. Then the source files of the project are not scanned for libraries, that need to be included and they need to be specified:
https://docs.platformio.org/en/latest/librarymanager/ldf.html#dependency-finder-mode

PR #653, opened 2020-01-09
Use recursive config patching

codefrey:
* More flexible patching
* Better readable code


codefrey:
This PR is a rebased and updated copy of #628.

The behaviour of the patching has now slightly changed. When an element in the patch JSON has a value of `null`, the element with the same name in the config will now be removed, instead of being set to `null`.

codefrey:
Had to force push. Somehow the commit still had the old code.

PR #659, opened 2020-02-01
Added deep sleep functionality for ESP32

EinfachArne:
I added the deep sleep functionality for ESP32 according to the documentation: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/sleep_modes.html


luebbe:
> 
> 
> Code looks good to me but I am not able to test this, as I don't have any ESP32 device.

Same here...

labodj:
Just tested on ESP32, it seems that it works.

Issue #648, opened 2020-01-02
OTA with MQTT TLS not working


nemidiy:
Hi everyone, happy new year!

I finally got to the point where I could test OTA updates under TLS. As we know it does not work.
Doc gives a hint :

> ASYNC_TCP_SSL_ENABLED
> This compiler flag allows to use SSL encryption for MQTT connections. All other network 
> connections still can not be encrypted like HTTP or OTA.

What was rather painful to find is that for esp32 [AsyncTcp](https://github.com/me-no-dev/AsyncTCP) has no support for TLS. There is a PR though that includes client side TLS using an mbed lib. Since all this adds an extra level of complexity I-ll start by debugging on ESP8266 and then if I can make that work Ill see what I can do in esp32 (my board of choice).

If anyone has any ideas on what the problem is for OTA not working on TLS please let me know :)

Here is what I am doing :
```shell
python ota_updater.py -l mqtt.dc-iot.com -p 8883 -t devices/ -i testbox8266 --broker-tls-cacert ca.crt /home/nemi/workspace/dc/homie_test/firmware.bin
Connecting to mqtt broker mqtt.dc-iot.com on port 8883
Connected with result code 0
Waiting for device to come online...
Waiting for device info...
Publishing new firmware with checksum 367e1be9f437c66f747a54de50016720
Waiting for device info...
Expecting checksum 367e1be9f437c66f747a54de50016720, got fc9d1211fbc105e51d3e4359d76cd108, update failed!
```
mqtt.dc-iot.com resolves in my private DNS to my mqtt server, that-s not even a real domain.
I can see if I subscribe to the base topic (using mosquitto_sub) that the FW is pushed just fine but then when the board receives the message it just disconnects from the server

```shell
💡 Firmware test (1.0.0)
🔌 Booting into normal mode 🔌
SSL is: 1
Using fingerprint: 2362f4fbc7890e450d43bb377220e2ac85409936
{} Stored configuration
  • Hardware device ID: cc50e33ca8c5
  • Device ID: testbox8266
  • Name: Test Box 8266
  • Device Stats Interval: 10 sec
  • Wi-Fi: 
    ◦ SSID: dc-iot
    ◦ Password not shown
    ◦ IP: 192.168.0.201
    ◦ Mask: 255.255.255.0
    ◦ Gateway: 192.168.0.1
  • MQTT: 
    ◦ Host: mqtt.dc-iot.com
    ◦ Port: 8883
    ◦ SSL enabled: true
[...]

PR #660, opened 2020-02-02
(#648) Tweaks to avoid OTA halting when using TLS.

nemidiy:
+ During OTA do not publish any messages that are not OTA related
+ Publish 1 every 100 206 status messages

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -139,6 +139,7 @@ void BootNormal::loop() {
     return;
   }
 
+  if (_otaOngoing) return;
   // here, we finished the advertisement
 
   if (!_mqttConnectNotified) {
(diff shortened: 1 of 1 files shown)


nemidiy:
cpplint failed. I did a commit --amend but looks like someone with write privileges need to rerun that part of the pipeline.

codefrey:
@nemidiy The  force-pushed commit still contains the errors the linter is complaining about.

IMHO it would be better to rate limit the OTA calls to `_publishOtaStatus` in `__handleOTAUpdates` instead of just dismissing them.

nemidiy:
@codefrey yes, it was just the guy behind the keyboard messing up. You bring a valid point, I moved the code so that if anyone wants to come a up with a more sophisticated approach they don't have to move it. thanks!

kleini:
Will test the latest changes on my devices.

PR #669, opened 2020-03-21
Add supply voltage to stats (ESP8266 only)

arneki:
Send the current supply voltage to the `$stats/supply` topic. Updated `$stats` as well.

### DIFF
--- a/src/Homie/Boot/BootNormal.cpp
+++ b/src/Homie/Boot/BootNormal.cpp
@@ -2,6 +2,13 @@
 
 using namespace HomieInternals;
 
+// needed to read vcc
+#ifdef ESP32
+//FIXME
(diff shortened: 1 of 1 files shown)


arneki:
@luebbe I just discovered your awesome [node collection](https://github.com/luebbe/homie-node-collection) and realized, that the [adc node](https://github.com/luebbe/homie-node-collection/blob/master/src/AdcNode.cpp) does the same thing I've added here. Maybe it's better to keep the code of homie-esp8266 as clean as possible and do something like that... What do you think?

kleini:
I think, it would be better to have the ESP8266 voltage as a HomieNode instead to have it in the stats.

luebbe:
Closing, since we agreed to keep the core clean

PR #677, opened 2020-05-15
Add option to set retained config to /set topic

labodj:
It's useful when you have a device that doesn't need to restore it's state after a power loss or a reboot.
It's even more useful when you use this library in a device used as a "forwarder" for another device state.

For example I'm using this library in a ESP32 connected to a Controllino Maxi via Serial, so ESP will configure it's name and nodes in setup() and will act as a forwarder of states in loop(), I don't need the state to be restored upon power loss or reboot because Controllino has it's default state at boot, and it can even restore previous state if I want, ESP32 "virtual" state does not matter, controllino will always override it.

### DIFF
--- a/src/SendingPromise.cpp
+++ b/src/SendingPromise.cpp
@@ -7,6 +7,7 @@ SendingPromise::SendingPromise()
 , _property(nullptr)
 , _qos(0)
 , _retained(false)
+, _setRetained(true)
 , _overwriteSetter(false)
 , _range { .isRange = false, .index = 0 } {
 }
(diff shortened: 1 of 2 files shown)


labodj:
It's also should be noted that following [Homie v4 specifications](https://homieiot.github.io/specification/spec-core-develop/#property-command-topic)

> A Homie controller publishes to the set command topic with non-retained messages only.

The default behavior of this library is the opposite, and I know that this library is 3.0.0 compliant, but with my code you can let the user choose the default behavior.

euphi:
This was discussed for Homie V4. 

For reading the old state it was proposed to read the retained message of the property itself and not the set topic.

euphi:
See https://github.com/homieiot/convention/issues/150 and https://github.com/homieiot/convention/issues/51

Issue #682, opened 2020-06-16
Migrate from SPIFFS to LittleFS


luebbe:
Since SPIFFS is about to be deprecated soon and the compiler already throws a lot of warnings about migrating to spiffs, we should take this step soon.
I'll give it a shot, but I may need some help, since I'm not a C guru :)

PR #684, opened 2020-06-28
Change FS from SPIFFS to LittleFS

gorec2005:
Just replace FS ... work for me...

### DIFF
--- a/src/Homie/Config.cpp
+++ b/src/Homie/Config.cpp
@@ -4,34 +4,34 @@ using namespace HomieInternals;
 
 Config::Config()
   : _configStruct()
-  , _spiffsBegan(false)
+  , _littlefsBegan(false)
   , _valid(false) {
 }
 
-bool Config::_spiffsBegin() {
-  if (!_spiffsBegan) {
+bool Config::_littlefsBegin() {
+  if (!_littlefsBegan) {
 #ifdef ESP32
-    _spiffsBegan = SPIFFS.begin(true);
+    _littlefsBegan = LittleFS.begin(true);
 #elif defined(ESP8266)
-    _spiffsBegan = SPIFFS.begin();
+    _littlefsBegan = LittleFS.begin();
 #endif
-    if (!_spiffsBegan) Interface::get().getLogger() << F("✖ Cannot mount filesystem") << endl;
+    if (!_littlefsBegan) Interface::get().getLogger() << F("✖ Cannot mount filesystem") << endl;
   }
 
-  return _spiffsBegan;
+  return _littlefsBegan;
 }
 
 bool Config::load() {
-  if (!_spiffsBegin()) { return false; }
+  if (!_littlefsBegin()) { return false; }
 
   _valid = false;
 
-  if (!SPIFFS.exists(CONFIG_FILE_PATH)) {
+  if (!LittleFS.exists(CONFIG_FILE_PATH)) {
     Interface::get().getLogger() << F("✖ ") << CONFIG_FILE_PATH << F(" doesn't exist") << endl;
     return false;
   }
 
-  File configFile = SPIFFS.open(CONFIG_FILE_PATH, "r");
+  File configFile = LittleFS.open(CONFIG_FILE_PATH, "r");
   if (!configFile) {
     Interface::get().getLogger() << F("✖ Cannot open config file") << endl;
     return false;
@@ -148,7 +148,7 @@ bool Config::load() {
 }
 
 char* Config::getSafeConfigFile() const {
-  File configFile = SPIFFS.open(CONFIG_FILE_PATH, "r");
+  File configFile = LittleFS.open(CONFIG_FILE_PATH, "r");
   size_t configSize = configFile.size();
 
   char buf[MAX_JSON_CONFIG_FILE_SIZE];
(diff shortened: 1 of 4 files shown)


kleini:
I use

    pio run -t uploadfs

to upload my config.json for every device to SPIFFS. I don't use the config mode at all and I even use the compile flag to remove the code for the config mode. How do I upload the config.json to LittleFS?

luebbe:
I also replaced all occurrences of Spiffs with LittleFS, but I see two pending problems:
* Uploading to LittleFS is not yet supported by PlatformIO
* Other libraries on which Homie depends, also use Spiffs and should be migrated too (alt least EspAsyncWebServer)

I'm pretty sure we shouldn't mix the two and I'm also pretty sure that we should put everything into defines at least for some time, so that people can have spiffs as default. After a while we could make LittleFS the default.

Which is why I haven't pursued the issue further (yet)

kleini:
closes #682

gorec2005:
Hi! i got work it with this... https://github.com/me-no-dev/ESPAsyncWebServer/issues/792#issuecomment-648144145

MajorTwip:
Is LittleFS availlable on ESP32?

dalbert2:
Is there a migration strategy for moving from SPIFFS to LittleFS?  If SPIFFS is used in actual products, there is data stored in those products that can't just be thrown away if the software is upgraded.

luebbe:
That is exactly the reason why I haven't merged anything yet. I remember coming across a project that worked on in-place migration of the data, but can't find it now.

Maenthyx:
Hi,
I really enjoy homie-esp8266 since some days and would love to see this project to evolve. I wonder whether there are some news to this topic? May I add my2cents?
In my opinion, conversion of any existing filesystem/data on running systems is out-of-scope of this project. Adding code to support an automatic conversion from spiffs to littlefs (whatever this might look like) would result in (a lot of additional) code/dependencies which doesn't fit into the "Homie" scope. Have in mind that this would raise future efforts to maintain that "legacy code".
I also couldn't find any other project considering a conversion from spiffs to littlefs on running systems, so why should this project do?

Nevertheless, I feel the pain loosing data. So I'd suggest a process (and maybe tools/simple sketches) to download the data, update the system and upload the data. Either manually done, or by leveraging the OTA feature. Well documented, but not implemented into this project.

You also might want to read this thread, discussing this topic, options and efforts.
Start reading from "Feb 25 14:14": [Gitter Thread](https://gitter.im/esp8266/Arduino?at=5e551db3101adb4160c680f1)

  
  
Also, for a limited time and depending on the effort, both spiffs and littlefs could be supported by using compiler flags, resulting in running either spiffs or littlefs (but not both at the same time). So people have some time to migrate their systems.

Thoughts?

kleini:
Supporting both filesystems through compiler flags was an earlier suggestion, too. This absolutely makes sense. I personally am still missing a tool to upload config.json to LittleFS filesystem, so I can test it.

Maenthyx:
Afaik, ArduinoIDE supports uploading files to littlefs. But I'm not sure if it's already built in or a [plugin ](https://github.com/earlephilhower/arduino-esp8266littlefs-plugin)is required.
I use PlatformIO, and by simply setting littlefs in the environment configuration you can upload files to littlefs using the pio cli.

May I ask what tool you're missing? Don't you use ArduinoIDE or PlatformIO?

kleini:
I am using PlatformIO and did not have a look yet into PlatformIO and LittleFS. Good to hear, that it is supported.

src/Homie/Config.cpp
```diff
@@ -85,11 +85,11 @@ bool Config::load() {
   const char* reqWifiDns2 = reqWifi["dns2"] | "";
 
   uint16_t reqMqttPort = reqMqtt["port"] | DEFAULT_MQTT_PORT;
-  bool reqMqttSsl = reqMqtt["ssl"] | false;
+//  bool reqMqttSsl = reqMqtt["ssl"] | false;
```
kleini, line 88: This would break MQTT over SSL for me.
As this PR contains a mixture of multiple things:
- ota_updater changes
- LittleFS support
- broken SSL support
- changes on WiFi connect behaviour
we should split this up into multiple smaller PRs covering each a single topic. I will take care about a compiler flags driven switch between SPIFFS and LittleFS.

kleini (changes requested):

src/Homie/Config.cpp
luebbe: Thank you, for the review. Maybe the ota_updater changes are caused by merging my changes back into his fork?
I also started working on the LittleFS change but stopped when it came to the async web server, since it doesn't support LittleFS. IMO LittleFS can right now only be supported with HOMIE_CONFIG=0. When the web server is used, we should default to SPIFFS.

kleini: I did not have a look at the commits and commit history but I think, we should try to keep Git history a little bit clean. For me code changes are already hard to follow.
Maybe I can make compile fail if LittleFS is selected without disabling the web server.

luebbe: It's better with a clean start. If you give it a go using this PR as inspiration, we can close this one in a few days.
Failing the compilation should be easy with a set of nested #if(n)defs

PR #691, opened 2020-10-26
Fix missing "#if HOMIE_MDNS" clause inside BootNormal::_onWifiGotIp

thecky:
There's a missing #if clause, which prevent the homie build for an ESP environment with "-D HOMIE_MDNS=0".

This PR add the missing #if, so that a build with disabled mDNS is possible.
The #if clause is missing in these branches:
 - develop
 - master
 - v3.0

The PR is only for the develop-branch, but cherry-picking for the other branches should be possible.

Regards,
Thomas


kleini:
LGTM

luebbe:
Looks good to me ;)

PR #696, opened 2020-12-21
fix: Homie framework fails with exception when compiled with sloeber

WLimin:
Tested by NodeMCU, Esp8266 V2.7.4, Sensor & Event demo.

Signed-off-by: wang <w.limin@qq.com>

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -5,7 +5,8 @@ using namespace HomieInternals;
 HomieClass::HomieClass()
   : _setupCalled(false)
   , _firmwareSet(false)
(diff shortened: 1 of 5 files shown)


kleini:
I think, we need to get the CI fixed. before merging more stuff.

stritti (approved):

luebbe:
OK, I'll wait three more days for reviews to come in. If not, I'll merge the PR.

kleini:
Will have a look into this in the next days. But just understanding, what is going on, will take some hours.

PR #697, opened 2020-12-25
fix suggested parentheses warning

kleini:
although C++ operator precedence would be correct in this case


luebbe (approved):
LGTM I have the same fix in my fork 😃

PR #699, opened 2020-12-28
Support for LittleFS

kleini:
SPIFFS remains default and LittleFS needs to be explicitly selected with compiler flags. LittleFS can only be used if config mode is excluded.


kleini:
Does somebody have a pointer, how to get proper LittleFS support for ESP32? Arduino Core does not contain LittleFS for ESP32.

PR #707, opened 2021-04-22
Improve deep sleep documentation

gaetancollaud:


kleini (approved):
LGTM

luebbe:
I wasn't sure about the deepsleep parameter, because I never used it. Is it really microseconds? If yes then LGTM :)

gaetancollaud:
> I wasn't sure about the deepsleep parameter, because I never used it. Is it really microseconds? If yes then LGTM :)

Yes, it's in microseconds. I had to go into the source code to find it. That's why I wanted to improve the doc.

There is also a possibility to use a pin as an interrupt to wake up. But I didn't go too much into details. I just wanted to have a simple example with a timer.

PR #709, opened 2021-05-10
Add information about HomieLoggerNode

CWempe:
Add information about HomieLoggerNode to the Community Projects.


CWempe:
see https://github.com/homieiot/homie-esp8266/issues/708

Issue #713, opened 2021-07-03
Little fix for DEBUG build flag


elbowz:
Activating `DEBUG` build flag I get the error:
```
.pio/libdeps/nodemcuv2/Homie/src/Homie/Boot/BootNormal.cpp: In member function 'bool HomieInternals::BootNormal::__handleNodeProperty(char*, char*, const AsyncMqttClientMessageProperties&, size_t, size_t, size_t)':
.pio/libdeps/nodemcuv2/Homie/src/Homie/Boot/BootNormal.cpp:1117:74: error: 'homieNode' was not declared in this scope
     Interface::get().getLogger() << F("Recived network message for ") << homieNode->getId() << endl;
```

Simply move the block:
https://github.com/homieiot/homie-esp8266/blob/9cd83972f27b394eab8a5e3e2baef20eea1b5408/src/Homie/Boot/BootNormal.cpp#L1125-L1127 

under:
https://github.com/homieiot/homie-esp8266/blob/9cd83972f27b394eab8a5e3e2baef20eea1b5408/src/Homie/Boot/BootNormal.cpp#L1151 

or viceversa.

Always thanks for this wonderful project!

elbowz:
Fixed with #719

PR #719, opened 2021-09-10
Add free heap to stats

elbowz:
Simply add to statistics  the value returned from `ESP.getFreeHeap()` (should works on ESP8266 and ESP32).
Maybe could be useful to understand if for some reason the heap memory fill along time.

Fix the little bug reported in #713


luebbe:
Thanks for your contribution

PR #721, opened 2021-09-13
Homie.isConfigured() now check only if valid (not load json)

elbowz:
Avoid loading config each time want check if device is configured (calling `isConfigured()`).

Fix: Remove the spam message *"/homie/config.json doesn't exist"* in config mode, when put `isConfigured()` in `loop()`.

**notes:**
I have made some test and it seems work, but in this way you can call `Homie.isConfigured()` only after `Homie.setup()`.

...or maybe can be added something: 
```c++
bool HomieClass::isConfigured() {
  if(!_setupCalled) return Interface::get().getConfig().load();
  else return Interface::get().getConfig().isValid();
}
```

### DIFF
--- a/src/Homie.cpp
+++ b/src/Homie.cpp
@@ -130,8 +130,11 @@ void HomieClass::setup() {
     _selectedHomieBootMode = HomieBootMode::NORMAL;
   }
 
+  // load and check config file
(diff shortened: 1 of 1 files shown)


elbowz:
Are there some news, feedback or timing about this PR and the other one (#720)

Thx!

PR #735, opened 2022-09-30
Fix for Espressif32 platform > 3.5.0

labodj:
Signed-off-by: LaboDJ <2527836+labodj@users.noreply.github.com>


labodj:
@luebbe I don't know why CicleCI doesn't want to collaborate 😅 anyway this is the fix for newer Espressif platform, I just tested it

luebbe:
Thank you for the PR. I hope I'll find time to look into the problem. This weekend was crazy and the upcoming week doesn't look better.

